Issue #423, opened 2022-07-05
fix sample app


gogonzo:
sample app exploratory fails because one of the server_args is not named.

pawelru:
This was fixed

PR #422, opened 2022-07-04
add reporter to modules

mhallal1:
closes #386 
dependency: https://github.com/insightsengineering/teal.reporter/pull/104

-Added reporter into all modules except `tm_front_page`, `tm_file_viewer` and `tm_data_table` as I did not see the necessity there, let me know if you disagree.
-`tm_scatterplot_matrix` plots were displayd as blank in the report, issue opened in `teal.reporter`: https://github.com/insightsengineering/teal.reporter/issues/98
-The bottom outliers table in the `tm_outliers` app is not rendered in the chunks therefore it is not included in the report (only the upper table and the plot).
-the reporter is also added in `tm_variable_browser` to add the plot, i am open to remove it if there is no need for it.

Please test with the example apps of each module

PR #424, opened 2022-07-05
fix sample app

gogonzo:
Closes https://github.com/insightsengineering/teal.modules.general/issues/423


gogonzo:
will be addressed here https://github.com/insightsengineering/teal.modules.general/pull/422

PR #430, opened 2022-07-22
reporter cleaning

mhallal1:
part of https://github.com/insightsengineering/teal.reporter/issues/88, https://github.com/insightsengineering/teal.reporter/issues/120 and https://github.com/insightsengineering/teal.slice/issues/56

- Use UI and SRV from simple_reporter
- Remove adding title for `Filter state` and `Show R code`

### DIFF
--- a/R/tm_t_crosstable.R
+++ b/R/tm_t_crosstable.R
@@ -129,7 +129,6 @@ ui_t_crosstable <- function(id, datasets, x, y, show_percentage, show_total, pre
     encoding = div(
       ### Reporter
       teal.reporter::simple_reporter_ui(ns("simple_reporter")),
-      shiny::tags$br(),
       ###
       tags$label("Encodings", class = "text-primary"),
       teal.transform::datanames_input(list(x, y)),
--- a/R/tm_t_crosstable.R
+++ b/R/tm_t_crosstable.R
@@ -128,11 +128,7 @@ ui_t_crosstable <- function(id, datasets, x, y, show_percentage, show_total, pre
     ),
     encoding = div(
       ### Reporter
-      shiny::tags$div(
-        teal.reporter::add_card_button_ui(ns("addReportCard")),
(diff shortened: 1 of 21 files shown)


R/tm_variable_browser.R
```diff
@@ -432,7 +427,6 @@ srv_variable_browser <- function(id, datasets, reporter, datasets_selected, ggpl
         card <- teal.reporter::TealReportCard$new()
         card$set_name("Variable Browser Plot")
         card$append_text("Variable Browser Plot", "header2")
-        card$append_text("Filter State", "header3")
```
nikolas-burkoff, line 435: Hmm so we need to record whether the use filtered data button was turned on or not as otherwise we could have the situation where say "Filtered State: x" but the Use filtered state button is set to FALSE and then the graph includes unfiltered data. Maybe we have `if (input$<<use_filter>>) card$append_fs(...)`

R/tm_t_crosstable.R
```diff
@@ -352,27 +347,22 @@ srv_t_crosstable <- function(id, datasets, reporter, label, x, y, basic_table_ar
         card <- teal.reporter::TealReportCard$new()
         card$set_name("Cross Table")
         card$append_text("Cross Table", "header2")
-        card$append_text("Filter State", "header3")
```
nikolas-burkoff, line 355: Do we have an issue to add encoding information into tmg modules as I could imagine say waning to know which join we've selected here

nikolas-burkoff (approved):
Just a few minor comments but this is good to be merged

R/tm_variable_browser.R
mhallal1: good point, updated.

R/tm_t_crosstable.R
mhallal1: https://github.com/insightsengineering/teal.modules.general/issues/434

PR #458, opened 2022-09-14
reduce imports

### LINKED ISSUE #456: Decrease dependencies
image

mhallal1:
closes #456 

-Checked all packages the are in `imports` with a single function call and removed them to `suggests` while using a check in the required module
-removed `magrittr` from `depends`
-replaced `digest::sha1` with `rlang::hash`

App deployed with this branch: https://rsconnect.roche.com/connect/#/apps/20589b31-da43-43b0-8206-8d0fb24bb385


R/tm_a_regression.R
```diff
@@ -91,6 +91,9 @@ tm_a_regression <- function(label = "Regression Analysis",
                             default_plot_type = 1,
                             default_outlier_label = "USUBJID") {
   logger::log_info("Initializing tm_a_regression")
+  if (!requireNamespace("methods", quietly = TRUE)) {
```
nikolas-burkoff, line 94: so methods is always available so we don't need this

mhallal1: updated.

R/tm_g_distribution.R
```diff
@@ -109,6 +109,18 @@ tm_g_distribution <- function(label = "Distribution Module",
                               pre_output = NULL,
                               post_output = NULL) {
   logger::log_info("Initializing tm_g_distribution")
+  if (!requireNamespace("ggpmisc", quietly = TRUE)) {
```
nikolas-burkoff, line 112: Also need broom ?

image

nikolas-burkoff (commented):
I would probably add a general update to the NEWS for this as well

R/tm_g_distribution.R
mhallal1: done.

nikolas-burkoff (approved):
👍

PR #460, opened 2022-09-16
remove hard coded ADSL

### LINKED ISSUE #459: Fix hard coded ADSL on refactor branch
We should not have hard-coded ADSL value in the module:

https://github.com/insightsengineering/teal.modules.general/blob/c5a2c8bae45c8a01b2704c6b799b708b1f99f9c9/R/tm_variable_browser.R#L1085

Also: Don't think this is right https://github.com/insightsengineering/teal.modules.general/blob/c5a2c8bae45c8a01b2704c6b799b708b1f99f9c9/R/tm_variable_browser.R#L132 as having join_keys is independent of having parent

Either:
1) Remove the functionality to hide ADSL variables for child datasets in the variable_browser 

or 

2) Add an argument to tm_variable_browser which is `parent` which defaults to "ADSL" like tmc modules do. Then if parent is in names(data) then the checkbox for showing parent variables or not is shown (as then for iris we don't see that)
and then the "ADSL" can be replaced with parent

nikolas-burkoff:
Closes #459 

Test with:

```
library(teal.modules.general)
library(scda)

ADSL <- synthetic_cdisc_data("latest")$adsl
ADTTE <- synthetic_cdisc_data("latest")$adtte

app <- init(
  data = cdisc_data(
    cdisc_dataset("ADSL", ADSL), cdisc_dataset("ADTTE", ADTTE),check = FALSE
  ),
  modules(
    tm_variable_browser()
  )
)

shinyApp(app$ui, app$server)

app <- init(
  data = teal_data(
    dataset("IRIS", iris),
    check = FALSE
  ),
  modules(
    tm_variable_browser()
  )
)

shinyApp(app$ui, app$server)


app <- init(
  data = teal_data(
    dataset("IRIS", iris),
    check = FALSE
  ),
  modules(
    tm_variable_browser(parent_dataname = character(0))
  )
)

shinyApp(app$ui, app$server)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,7 @@
 * Added the `teal.reporter` functionality to all modules.
 * Added log transformation options to `tm_g_scatterplot`.
 * Added `server_rendering` flag to `tm_data_table` to control whether the table is rendered server or client side.
+* Added `parent_dataname` argument to `tm_variable_browser` to allow specification of parent dataset for this module.
 
 ### Bug fixes
 * Fixed an overflow for very wide `tm_a_pca` tables.
--- a/R/tm_variable_browser.R
+++ b/R/tm_variable_browser.R
@@ -9,6 +9,11 @@
 #'
 #' @inheritParams teal::module
 #' @inheritParams shared_params
+#' @param parent_dataname (`character`) If this dataname exists in `datasets_selected`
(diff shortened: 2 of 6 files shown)


mhallal1 (approved):
as expected.

Issue #312, opened 2022-01-20
[UAT] scale text on the variable browser plot with the plot hieght and width


kpagacz:
The text in the right upper corner has a fixed size. As a consequence, it's possible to open the plot with the modal button, increase the plot size and get a very small text (almost ineligible) like this:
![image](https://user-images.githubusercontent.com/24480246/150348622-150d48f5-d693-4848-ab23-8107e939ccd2.png)

Downloading the file produces a gigantic plot with a tiny text: 
![image](https://user-images.githubusercontent.com/24480246/150348702-4d7d7e70-2d15-47c3-a149-c74ad37577a8.png)

Would be better if the text somehow scaled with the resolution.

chlebowa:
It's easy enough to do for a single `ggplot` but nested `gTree` objects are a different story.

nikolas-burkoff:
hmm - so it looks like the output from `plot_var_summary` is either a ggplot object (in which case we want to change the size of the text) or not - in which case it looks like this 
![image](https://user-images.githubusercontent.com/15201933/200891548-6d056a25-5f93-4c2e-9695-d257f281321f.png)
in which case we don't really care about the font size. 

So we could add the font size [inside this code block](https://github.com/insightsengineering/teal.modules.general/blob/main/R/tm_variable_browser.R#L945) - it's also probably easiest to have a font size slider in the UI to let users pick it rather than trying anything clever and automatic.

nikolas-burkoff:
You can also use the ggplot2_args to handle this - for example see [here](https://github.com/insightsengineering/teal.modules.general/blob/main/R/tm_a_pca.R#L425)   @Polkas is the person to ask about those and see teal.widgets for more details

lcd2yyz:
I do like the idea of font size slider, it would be a nice value added to all plot rendering through the widget. 
Potential follow-up question
- How to determine/set a font size range that make sense?
- Does the font size adjustment need to be handled differently between `ggplot` vs `grid/gTree` objects?

nikolas-burkoff:
> I do like the idea of font size slider, it would be a nice value added to all plot rendering through the widget.

So this is a nice idea but I'm not sure it's feasible as `plot_with_settings` allows lots of different types of graphical output to be displayed all of which handle font sizes differently (if you look at the km plot in tmc or the lineplot in teal.goshawk or the scatterplotmatrix in tmg) and I think different parts of graphs (axes, titles etc.) can be controlled differently too. 

[...]

PR #489, opened 2022-11-10
example fontsize slider

nikolas-burkoff:
Related to https://github.com/insightsengineering/teal.modules.general/issues/312

This is a rough and ready way of solving the issue - some work is needed for the UI (and maybe the slider should be hidden - with checks on input$font_size not being NULL etc.

image
image


nikolas-burkoff:
I haven't had much experience with ggplot2_args but looking at the pca example instead of passing font_size as an explicit argument to `plot_var_summary` you can sort of add it in to the ggplot2_args directly but this isn't required

PR #490, opened 2022-11-16
[UAT] scale text on the variable browser plot with the plot hieght and width #312

chlebowa:
Closes [this issue](https://github.com/insightsengineering/teal.modules.general/issues/312).

Added collapsible slider to allow user to manipulate text size in plots.
Patched applying theme to plots.
Added aliases for module documentation.

### DIFF
--- a/R/tm_variable_browser.R
+++ b/R/tm_variable_browser.R
@@ -18,6 +18,14 @@
 #'   shown and in what order. Names in the vector have to correspond with datasets names.
 #'   If vector of length zero (default) then all datasets are shown.
 #'
+#' @aliases
+#'   tm_variable_browser_ui,
+#'   tm_variable_browser_srv,
+#'   tm_variable_browser,
+#'   variable_browser_ui,
+#'   variable_browser_srv,
+#'   variable_browser
+#'
 #'
 #' @export
 #'
@@ -104,6 +112,7 @@ ui_variable_browser <- function(id,
 
   shiny::tagList(
     include_css_files("custom"),
+    shinyjs::useShinyjs(),
     teal.widgets::standard_layout(
       output = fluidRow(
         htmlwidgets::getDependency("sparkline"), # needed for sparklines to work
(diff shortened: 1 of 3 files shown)


R/tm_variable_browser.R
```diff
@@ -156,7 +158,18 @@ ui_variable_browser <- function(id,
               } else {
                 shinyjs::hidden(x)
               }
-            }
+            },
+            # input user-defined text size
+            teal.widgets::panel_item(
+              title = "Plot settings",
+              collapsed = TRUE,
+              teal.widgets::optionalSliderInputValMinMax(
```
nikolas-burkoff, line 166: This is fine - I might have just used a plain shiny sliderInput

nikolas-burkoff (approved):
Looking good:

A few comments:
1) Please add a NEWS item
2) I wonder if we should move the ui to here?
image - I'll leave that up to you
3) I wonder how much effort it is to add the "theme" option like we do in the other tmg modules - probably not worth adding tbh unless it's super easy

R/tm_variable_browser.R
```diff
@@ -104,6 +112,7 @@ ui_variable_browser <- function(id,
 
   shiny::tagList(
     include_css_files("custom"),
+    shinyjs::useShinyjs(),
```
nikolas-burkoff, line 115: So no harm putting this here but it is called [here](https://github.com/insightsengineering/teal/blob/eb373dfdfb06aaabc97a10b7c5c7ce082d6429e3/R/include_css_js.R#L80) from [here](https://github.com/insightsengineering/teal/blob/79678e78267098d1a0377f0b6b9cd8bba1cea988/R/module_teal.R#L95)

chlebowa:
> I wonder how much effort it is to add the "theme" option like we do in the other tmg modules - probably not worth adding tbh unless it's super easy

That would require rendering the UI for plot settings server side as a default theme may be supplied by `ggplot2_args`. Not too hard, but not super easy either.

nikolas-burkoff (approved):

R/tm_variable_browser.R
chlebowa: Ah, I was wondering whether `shinyjs` is added to all apps by default.

chlebowa:
>> I wonder how much effort it is to add the "theme" option like we do in the other tmg modules - probably not worth adding tbh unless it's super easy

> That would require rendering the UI for plot settings server side as a default theme may be supplied by ggplot2_args. Not too hard, but not super easy either.

Actually, no it wouldn't, as a theme is never passed as a complete theme, but a set of element definitions in `ggplot2_args`. It's easy enough.

But I ran into another problem there, an assertion fails that probably shouldn't. Consulting with @Polkas .

chlebowa:
Adding theme selection had two unexpected effects:
1) revealed problem with argument checks in `resolve_ggplot2_args`, described in detail [here](https://github.com/insightsengineering/teal.widgets/issues/108)
2) overwrote the dev-specified x label rotation of 45 degrees; I added another input for that to the UI so that user settings take precedence before theme

R/tm_variable_browser.R
```diff
@@ -246,6 +264,24 @@ srv_variable_browser <- function(id,
...
+    # add used-defined text size to ggplot arguments passed from caller frame
+    all_ggplot2_args <- reactive({
+      user_text <- teal.widgets::ggplot2_args(
+        theme = list("text" = ggplot2::element_text(size = input[["font_size"]]),
+                     "axis.text.x" = ggplot2::element_text(angle = input[["label_rotation"]], hjust = 1))
+      )
+      user_theme <- getFromNamespace(sprintf("theme_%s", input[["ggplot_theme"]]), ns = "ggplot2")
+      user_theme <- user_theme()
+      # temporary fix to circumvent assertion issue with resolve_ggplot2_args
+      # drop problematic elements
```
nikolas-burkoff, line 276: https://github.com/insightsengineering/teal.widgets/issues/108

R/tm_variable_browser.R
```diff
@@ -175,6 +173,21 @@ ui_variable_browser <- function(id,
...
+              title = "Plot settings",
+              collapsed = TRUE,
+              selectInput(inputId = ns("ggplot_theme"), label = "ggplot2 theme",
+                          choices = c("gray", "bw", "linedraw", "light",
+                                      "dark", "minimal", "classic", "void", "test"),
+                          selected = "grey"),
+              fluidRow(
+                column(6, sliderInput(inputId = ns("font_size"), label = "font size",
+                                      min = 5L, max = 30L, value = 15L, step = 1L, ticks = FALSE)),
+                column(6, sliderInput(inputId = ns("label_rotation"), label = "rorate x labels",
```
nikolas-burkoff, line 187:
```suggestion
                column(6, sliderInput(inputId = ns("label_rotation"), label = "rotate x labels",
```

Issue #376, opened 2022-04-07
scatterplot: Bad error message when row facet and column facet are the same variable


denisovan31415:
![Screen Shot 2022-04-07 at 1 39 17 PM](https://user-images.githubusercontent.com/12874196/162273983-0fc88dd4-3b6f-451c-b827-3c74f93e4095.png)

donyunardi:
### Acceptance Criteria: 
Provide a more user-friendly error message

BLAZEWIM:
yes, thanks for noticing

PR #494, opened 2022-11-22
376 facet error@main

BLAZEWIM:
Fixes #[376](https://github.com/insightsengineering/teal.modules.general/issues/376)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,9 @@
 * Improved UI labels and plot panel title in `tm_g_association`.
 * Added inputs `tm_variable_browser` module for text size and plot theme.
 
+### Bug fixes
+* Fixed a bug in `tm_g_scatterplot` when selected x and y facets were the same.
+
 # teal.modules.general 0.2.15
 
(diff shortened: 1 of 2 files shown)


chlebowa:
Looks good to me, however, when I choose a numeric value for facetting, I get 
```
`Row facetting` variable  must be of class `character`, `factor`, `Date`, or `integer`.
```
which is not really a good error message, the end user may find it confusing.

Continuous variables should not be offered for faceting or the message should be better.
Perhaps create another issue to address this?

mhallal1 (approved):

BLAZEWIM:
> Looks good to me, however, when I choose a numeric value for facetting, I get
> 
> ```
> `Row facetting` variable  must be of class `character`, `factor`, `Date`, or `integer`.
> ```
> 
> which is not really a good error message, the end user may find it confusing.
> 
> Continuous variables should not be offered for faceting or the message should be better. Perhaps create another issue to address this?

I propose to add it during UAT

PR #499, opened 2022-12-23
modules by AC

chlebowa:
some more modules


R/tm_a_pca.R
```diff
@@ -269,10 +269,40 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
+      select_validation_rule = list(
+        dat = ~if (length(.) < 2L) "Please select more than 1 variable to perform PCA.",
+        response = shinyvalidate::compose_rules(
+          shinyvalidate::sv_optional(),
+          rule_diff("dat")
+        )
+      )
+    )
+
+    rule_dupl <- function(value) {
```
nikolas-burkoff, line 293: So you're not actually using value here, which although is ok, maybe isn't that clear?

R/tm_a_pca.R
```diff
@@ -290,26 +320,22 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
     computation <- reactive({
       req(merged$anl_q_r())
       # inputs
-      keep_cols <- as.character(merged$anl_input_r()$columns_source$dat)
+      keep_cols <- selector_list()$dat()$select
```
nikolas-burkoff, line 323: as discussed, we should change these back to merged$anl_input_r() throughout -> and maybe we should add an issue into teal.transform to give us helper functions to access these things

R/tm_a_pca.R
```diff
@@ -290,26 +320,22 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
       ANL <- merged$anl_q_r()[["ANL"]] # nolint
 
       # inputs validation
-      validate(need(length(keep_cols) > 1, "Please select more than 1 variable to perform PCA."))
-      teal::validate_has_elements(keep_cols, "Please select columns")
-      validate(need(
-        all(vapply(ANL[keep_cols], function(x) is.numeric(x) && all(!is.infinite(x)), logical(1))),
-        "PCA is only defined for (finite) numeric columns."
-      ))
+      teal::validate_inputs(iv_r())
```
nikolas-burkoff, line 331: could this be moved to top of reactive?

R/tm_a_pca.R
```diff
@@ -876,20 +877,50 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
-      } else if (input$plot_type == "Eigenvector plot") {
-        plot_pc_var(computation())
-      } else {
-        stop("Unknown plot")
-      }
+      iv_extra <- shinyvalidate::InputValidator$new()
+      iv_extra$add_rule("ggtheme", shinyvalidate::sv_required("Please select a theme."))
+
+      iv_elbow <- shinyvalidate::InputValidator$new()
+      iv_elbow$condition(~ input$plot_type == "Elbow plot")
```
nikolas-burkoff, line 884: As you are doing this inside the reactive then you already know what input$plot_type is so you only need to create the one validator for that specific type rather than creating 4

R/tm_a_regression.R
```diff
@@ -241,24 +249,66 @@ srv_a_regression <- function(id,
   with_filter <- !missing(filter_panel_api) && inherits(filter_panel_api, "FilterPanelAPI")
   checkmate::assert_class(data, "tdata")
   moduleServer(id, function(input, output, session) {
-    anl_merged_input <- teal.transform::merge_expression_module(
+
+    rule_rvr1 <- function(value) {
+      if (isTRUE(input$plot_type == "Response vs Regressor")) {
+        if (length(value) > 1L)
+          "This plot can only have one regressor."
```
nikolas-burkoff, line 256: So I think highlighting this ui is misleading:
image

as it is a valid regression -> if anything we should highlight the radiobutton

R/tm_a_pca.R
```diff
@@ -290,26 +320,22 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
       standardization <- input$standardization
       center <- standardization %in% c("center", "center_scale") # nolint
       scale <- standardization == "center_scale"
       ANL <- merged$anl_q_r()[["ANL"]] # nolint
 
       # inputs validation
-      validate(need(length(keep_cols) > 1, "Please select more than 1 variable to perform PCA."))
-      teal::validate_has_elements(keep_cols, "Please select columns")
-      validate(need(
-        all(vapply(ANL[keep_cols], function(x) is.numeric(x) && all(!is.infinite(x)), logical(1))),
```
nikolas-burkoff, line 304: do we still need this ?

R/tm_a_regression.R
```diff
@@ -178,30 +178,38 @@ ui_a_regression <- function(id, ...) {
...
-        ns("label_var"),
-        multiple = FALSE,
-        label = "Outlier label"
-      )),
+          ),
+          min = 1, max = 10, value = 9, ticks = FALSE, step = .1
+        )
+      ),
+      conditionalPanel(
+        condition = "input['show_outlier']",
```
nikolas-burkoff, line 205: as this is the same condition, couldn't it be moved into the conditionalPanel above?

R/tm_a_regression.R
```diff
@@ -843,6 +856,9 @@ srv_a_regression <- function(id,
     )
 
     output$text <- renderText({
+      req(iv_r()$is_valid())
```
nikolas-burkoff, line 859: so the way this is done, you now don't get all the things that are wrong shown:
image

and I'd argue the theme choice is much less important than you haven't selected a variable

R/tm_a_regression.R
nikolas-burkoff: similarly this one:
image

although unlike the case above, in this case I do wonder why we allow this for other plot types...

nikolas-burkoff (commented):
Looking good overall some comments in line

chlebowa:
#503

R/tm_a_regression.R
chlebowa: > So I think highlighting this ui is misleading:
> (...)
> as it is a valid regression -> if anything we should highlight the radiobutton

This is the original message:
"Response vs Regressor plot is provided only for regressions with exactly one regressor.\n",
            "Choose another plot or reduce number of regressors"

If anything, we should highlight both.

I don't think printing all of that 👆 in input widgets is a good idea.

R/tm_a_pca.R
```diff
@@ -269,10 +269,40 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
       )
     }
 
-    anl_merged_input <- teal.transform::merge_expression_module(
+
+    rule_diff <- function(other) {
```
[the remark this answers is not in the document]
chlebowa: You mean you prefer this?
```
~ if (isTRUE(is.element(., selector_list()$dat()$select))) "Response must not have been used for PCA."
```

R/tm_a_pca.R
```diff
@@ -876,20 +877,50 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
+             "Elbow plot" = plot_elbow(computation()),
+             "Circle plot" = plot_circle(computation()),
+             "Biplot" = plot_biplot(computation()),
+             "Eigenvector plot" = plot_pc_var(computation()),
+             stop("Unknown plot"))
     })
 
-    plot_r <- reactive(output_q()[["g"]])
+    plot_r <- reactive({
+      output_q()[["g"]]
```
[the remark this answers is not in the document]
chlebowa: These are three independent output elements and only the bottom one is always displayed.

R/crule.R
```diff
@@ -0,0 +1,66 @@
+#' add condition to rule function
```
mhallal1, line 1:
```suggestion
#' Add condition to rule function
```

R/crule.R
```diff
@@ -0,0 +1,66 @@
...
+#' iv$add_rule("id", crule(~ if(!. %in% set), !is.null(set)))
+#' }
+#'
+#' @export
+#'
+crule <- function(rule, condition, ...) {
+  checkmate::assert(
+    checkmate::check_class(rule, "function"),
+    checkmate::check_class(rule, "formula")
+  )
```
mhallal1, line 52: `checkmate::assert_multi_class()` might solve the double checks here.

R/tm_a_pca.R
```diff
@@ -269,10 +262,58 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
+    iv_r <- reactive({
+      iv <- shinyvalidate::InputValidator$new()
+      teal.transform::compose_and_enable_validators(iv, selector_list)
+    })
+
+    iv_extra <- reactive({
+      iv_extra <- shinyvalidate::InputValidator$new()
+      if (input$plot_type %in% c("Circle plot", "Biplot")) {
+        iv_extra$add_rule("x_axis", shinyvalidate::sv_required("Need X axis"))
+        iv_extra$add_rule("y_axis", shinyvalidate::sv_required("Need Y axis"))
```
mhallal1: unselecting any of the 2 above is rendering warnings in the console:
```
Warning: Error in : Problem while computing aesthetics.
ℹ Error occurred in the 1st layer.
Caused by error in `FUN()`:
! object 'PC2' not found 
 when evaluating qenv code:
```

R/tm_a_pca.R
```diff
@@ -269,10 +262,58 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
+        iv_extra$add_rule("x_axis", rule_dupl)
+        iv_extra$add_rule("y_axis", rule_dupl)
+      }
+      if (input$plot_type == "Circle plot") {
+        iv_extra$add_rule("variables", shinyvalidate::sv_required("Need Variables"))
+      }
+      if (input$plot_type == "Eigenvector plot") {
+        iv_extra$add_rule("pc", shinyvalidate::sv_required("Need PC"))
```
mhallal1: unselecting this breaks the app.

R/tm_a_pca.R
```diff
@@ -269,10 +262,53 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
+      teal.transform::compose_and_enable_validators(iv, selector_list)
+    })
+
+    iv_extra <- shinyvalidate::InputValidator$new()
+    iv_extra$add_rule("x_axis", crule(~ if (!shinyvalidate::input_provided(.)) "Need X axis",
+                                      isTRUE(input$plot_type %in% c("Circle plot", "Biplot"))))
+    iv_extra$add_rule("y_axis", crule(~ if (!shinyvalidate::input_provided(.)) "Need Y axis",
+                                      isTRUE(input$plot_type %in% c("Circle plot", "Biplot"))))
+    iv_extra$add_rule("x_axis", crule(rule_dupl, isTRUE(input$plot_type %in% c("Circle plot", "Biplot"))))
+    iv_extra$add_rule("y_axis", crule(rule_dupl, isTRUE(input$plot_type %in% c("Circle plot", "Biplot"))))
```
mhallal1, line 300: There are warnings when these are deselected:
```
Warning: Error in : Can't subset columns that don't exist.
✖ Column `PC1` doesn't exist. 
 when evaluating qenv code:
```

R/tm_a_pca.R
```diff
@@ -269,10 +262,53 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
+    iv_extra$add_rule("variables", crule(~ if (!shinyvalidate::input_provided(.)) "Need Variables",
+                                         identical(input$plot_type, "Circle plot")))
+    iv_extra$add_rule("pc", crule(~ if (!shinyvalidate::input_provided(.)) "Need PC",
+                                  identical(input$plot_type, "Eigenvector plot")))
```
mhallal1, line 304: Deselecting this breaks the app.

R/tm_a_regression.R
```diff
@@ -241,20 +245,61 @@ srv_a_regression <- function(id,
...
+    })
+
+    iv_out <- shinyvalidate::InputValidator$new()
+    iv_out$condition(~ isTRUE(input$show_outlier))
+    iv_out$add_rule("label_var", shinyvalidate::sv_required("Please provide an `Outlier label` variable"))
+    iv_out$enable()
+
+    iv_theme <- shinyvalidate::InputValidator$new()
+    iv_theme$add_rule("ggtheme", shinyvalidate::sv_required("Please select a theme."))
+    iv_theme$enable()
```
mhallal1, line 293: why does this need a separate `iv?` can we include it in `iv_r`?

R/tm_g_distribution.R
```diff
@@ -343,9 +336,84 @@ srv_distribution <- function(id,
...
+    )
+
+    iv_r <- reactive({
+      iv <- shinyvalidate::InputValidator$new()
+      teal.transform::compose_and_enable_validators(iv, selector_list)
+    })
+
+    iv_theme <- shinyvalidate::InputValidator$new()
+    iv_theme$add_rule("ggtheme", shinyvalidate::sv_required("Please select a theme."))
+    iv_theme$enable()
```
mhallal1, line 382: same here.

R/tm_g_distribution.R
```diff
@@ -343,9 +336,84 @@ srv_distribution <- function(id,
...
+        strata_i = strata_var,
+        group_i = group_var
+      ),
+      data,
+      select_validation_rule = list(
+        dist_i = shinyvalidate::sv_required("Please select a variable")
+      ),
+      filter_validation_rule = list(
+        strata_i = shinyvalidate::compose_rules(
+          crule(rule_req, isTRUE(input$dist_tests %in% dists1)),
```
mhallal1, line 368: stratification levels filter is always required when unselected, i think that the condition here is not correct.

R/tm_g_distribution.R
```diff
@@ -343,9 +336,84 @@ srv_distribution <- function(id,
...
+          crule(rule_dupl, identical(input$dist_tests, "Fligner-Killeen"))
```
mhallal1, line 369: this is also applicable for the test part only and not the plot, check it on main.

mhallal1: be aware of the tests applied to the distinct parts: displot, qqplot and tests

R/crule.R
```diff
@@ -0,0 +1,66 @@
...
+#'
+#' iv <- InputValidator$new()
+#' iv$add_rule("id", sv_required())
+#' iv$add_rule("id", crule(custom_rule, !is.null(set)))
+#' iv$add_rule("id", crule(~ if(!. %in% set), !is.null(set)))
+#' }
+#'
+#' @export
+#'
+crule <- function(rule, condition, ...) {
```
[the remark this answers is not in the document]
chlebowa: This is a temporary address, I imagine it would go to `teal` like `validate_inputs`.

R/crule.R
```diff
@@ -0,0 +1,66 @@
...
+    rule <- rlang::as_function(rule)
+  }
+  # add backstop for Error in str2lang(s) : argument must be a character string resulting from... line break in console???
+
+  ex <- list(
+    str2lang(sprintf("if (isFALSE(%s)) return(NULL)", deparse(substitute(condition)))),
+    body(rule)
+  )
+  body(rule) <- as.call(c(as.name("{"), ex))
+  rule
```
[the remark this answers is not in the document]
chlebowa: No, because a whole function must be returned, not just the body.

R/tm_g_distribution.R
chlebowa: This one is actually necessary.

R/crule.R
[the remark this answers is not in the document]
mhallal1: so let us move it into `utilities.R` until it gets moved into `teal`.

R/crule.R
mhallal1: ??

R/tm_a_pca.R
```diff
@@ -938,6 +946,20 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
+                                            "Plot settings are required" = iv_extra))
+      validation()
+      tags$div(
+        class = "overflow-scroll",
+        uiOutput(session$ns("tbl_importance_ui")),
+        hr(),
+        uiOutput(session$ns("tbl_eigenvector_ui")),
+        hr(),
+        teal.widgets::plot_with_settings_ui(id = session$ns("pca_plot"))
+      )
```
mhallal1, line 960: this should stay in the ui section.

chlebowa: Then how duplication of validation messages is avoided.

mhallal1 (approved):

R/utils.R
```diff
@@ -351,3 +351,69 @@ is_tab_active_js <- function(id, name) {
...
+#' @export
+#'
+crule <- function(rule, condition, ...) {
+  checkmate::assert_multi_class(rule, c("function", "formula"))
+  checkmate::assert_class(substitute(condition), "call")
+
+  if (inherits(rule, "formula")) {
+    rule <- rlang::as_function(rule)
+  }
+  # add backstop for Error in str2lang(s) : argument must be a character string resulting from... line break in console???
```
mhallal1, line 411: please update this message.

chlebowa: fixed

Issue #521, opened 2023-03-24
Review `tmg` documentation before release


donyunardi:
### Summary
Related with insightsengineering/nestdevs-tasks#25
In preparation to release, we need to review/assess all documentation in the package. 

### Acceptance Criteria
* [x] Review roxygen examples on all functions
  * Test examples
  * If it's using chunk, update to `qenv`.
* [x] Review vignettes on these topics only:
  * Review any content around `qenv`
  * Review any content around `chunks` (i.e. make sure the right language around deprecation)
  * Review any content around shinyvalidate
* [x] redo the content of WORDLIST
  * For instruction on how to do this, visit https://github.com/insightsengineering/coredev-tasks/issues/235
  * Check the PR for it https://github.com/insightsengineering/teal.modules.general/pull/528
* [x] Update NEWS
* [x] Update README (if applicable)
* [x] Review DESCRIPTION file (i.e check dependency versions, etc)
  * Be sure to update teal.code dependency to be >= 0.3.0

PR #485, opened 2022-10-27
Teal refactor

nikolas-burkoff:
Part of https://github.com/insightsengineering/teal/issues/731

PR #530, opened 2023-06-26
521 documentation review before the release

m7pr:
Based on specification from https://github.com/insightsengineering/teal.modules.general/issues/521

I checked that this PR https://github.com/insightsengineering/teal.modules.general/pull/485/ replaced all `teal.code::chunks_*` with `teal.code::eval_code` or `teal.code::new_qenv`. This looks like backend functionality (not exposed to the user) so `chunks` were not exposed in examples, hence even though I reviewed examples, nothing was there to be checked. Vignettes, nor README nor DESCRIPTION never mentioned the usage of `chunks` so no need to update them with any deprecation notes.

Unsure if there is a bigger purpose of this task besides bumping teal.code in DESCRIPTION file to 0.3.0. Am I missing something @gogonzo ?


gogonzo:
> backend functionality (not exposed to the user) so `chunks` were not exposed in examples, hence even though I reviewed examples, nothing was there to be checked. Vignettes, nor README nor DESCRIPTION never mentioned the usage of `chunks` so no need to update them with any deprecation notes.

If you reviewwed examples please make sure you actually run them in an interactive session as they are not covered by `devtools::run_examples`.


Unsure if there is a bigger purpose of this task besides bumping teal.code in DESCRIPTION file to 0.3.0. Am I missing something @gogonzo ?

Yes, change is massive. We changed argument `datasets` (FilteredData) to `data` (`tdata` - list of reactive data + code as attribute). 

changes main vs last release https://github.com/insightsengineering/teal.modules.general/compare/2022_10_13...main?diff=split

m7pr:
@gogonzo I did check that there was a lot of changes during this PR https://github.com/insightsengineering/teal.modules.general/pull/485
However I run all examples manually as they have this interactive statement verification

```{R}
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```

Some needed extra packages to be installed, like `install.packages(c('goftest', 'ggpmisc'))` at `tm_g_distribution.R` or `install.packages('ggExtra')` at `tm_g_scatterplot.R`, or `install.packages('sparkline')` at `tm_variable_browser.R` but those are listed in Suggests, so all is proper.

chlebowa:
> Some needed extra packages to be installed, like `install.packages(c('goftest', 'ggpmisc'))` at `tm_g_distribution.R` or `install.packages('ggExtra')` at `tm_g_scatterplot.R`, or `install.packages('sparkline')` at `tm_variable_browser.R` but those are listed in Suggests, so all is proper.

Did you also add appropriate messages for installing those packages? Ones like this appear in several places in `teal.modules.general`.
```
if (!requireNamespace("rtables", quietly = TRUE)) {
  stop("Cannot load rtables - please install the package or restart your session.")
}
```

m7pr:
@chlebowa yes, those were already pasted as warnings/error messages

> Error in tm_variable_browser(label = "Variable browser", ggplot2_args = teal.widgets::ggplot2_args(labs = list(subtitle = "Plot generated by Variable Browser Module")), :
Cannot load sparkline - please install the package or restart your session.

> Error in tm_g_scatterplot(label = "Scatterplot Choices", x = data_extract_spec(dataname = "ADSL", :
Cannot load package(s): ggExtra.
Install or restart your session.Cannot load package(s): colourpicker.
Install or restart your session.

> Error in tm_g_distribution(dist_var = data_extract_spec(dataname = "iris", :
Cannot load package(s): goftest.
Install or restart your session.

> Error in tm_g_distribution(dist_var = data_extract_spec(dataname = "iris", :
Cannot load package(s): ggpmisc.
Install or restart your session.Cannot load package(s): ggpp.
Install or restart your session.Cannot load package(s): goftest.
Install or restart your session.

gogonzo:
> Below example taken from `tm_g_scatterplot`

Since it is already there let's not change it, message is explicit and clear. I don't have an opinion yet how we should handle dependencies yet. This is something we'd need to discuss soon

chlebowa:
> paste(missing_packages, sep = ", ")))

Oh, good. Can you fix the message, though? This `sep` should be `collapse`. Or better yet, use `toString`.

m7pr:
We could open a prompt, asking a person for an action in this case

On a missing dependencies:
1) Do nothing and exit
2) Allow installation and exit
3) Allow installation and move forward with the function execution

m7pr:
> Oh, good. Can you fix the message, though? This sep should be collapse. Or better yet, use toString.

@chlebowa yeah, `toString` is neat https://github.com/insightsengineering/teal.modules.general/pull/530/commits/5387684609958ce5d28fef01f00b228f35167135

chlebowa:
> We could open a prompt, asking a person for an action in this case
> 
> On a missing dependencies:
> 
> 1. Do nothing and exit
> 2. Allow installation and exit
> 3. Allow installation and move forward with the function execution

It's not strictly missing, the packages are listed in `Suggests` for a reason (that eludes me). Let's keep it as is.

m7pr:
Yeah I know they are listed in Suggests. The popular convention is that, if a package is only used in one function that is not the main function of the package and package can live without this function then it's better to have the package in Suggests, so it's not installed by default when installing the initial package. The dependent packages in Suggests will be installed if someone specify `install.packages('initial package', dependencies = c("Depends", "Imports", "Suggests")`. Some packages are only used in vignettes or examples that are not run and those also should live in suggests.

The only question, that we could tackle on a separate discussion, is if we:
1) do not do anything with it
2) at least thrown a meaningful error for a person trying to use a function that is based on a Suggested package - this is what currently is happening
3) go one step beyond and also prepare a workaround that also allows to hit ENTER on a prompt so that missing packages get installed for the user and the function continues

chlebowa:
> Yeah I know they are listed in Suggests. The popular convention is that, **if** a package is only used in one function that is not the main function of the package and package can live without this function, **then** it's better to have the package in Suggests, so it's not installed by default when installing the initial package (so that installation time is shorter and there is less dependencies clutter). The dependent packages in Suggests will be installed if someone specify `install.packages('initial package', dependencies = c("Depends", "Imports", "Suggests"))`. Some packages are only used in vignettes or examples that are not run and those also should live in Suggests.

That is correct. The packages that we `Suggest` are most often ones we only use for examples.

> The only question, that we could tackle on a separate discussion, is if we:
> 
> 1. do not do anything with it

And that is what we do when they are encountered. Just raise an error.

m7pr:
alrighty, getting back to tmg review
- examples are run - no teal.code::chunks_*
- code is reviewed - no teal.code::chunks_*
- vignettes are checked - - teal.code::chunks_*
- no narration anywhere about teal.code::chunks_* nor teal.code::*qenv* 

I think it was a great job already made on this huge PR that introduced qenv

gogonzo (approved):
looks good. Thanks for updating `teal.code` depedency

m7pr:
Hey @gogonzo @pawelru the check for this PR does not pass because our own builds do not allow specific R CMD Check NOTES. In this case we do not allow any NOTES related to `"checking examples .* NOTE"` message https://github.com/insightsengineering/teal.modules.general/actions/runs/5401394235/jobs/9811839374?pr=530#step:36:50

In our case this is raised because examples take more than 5 seconds - they actually run for 8 seconds 
https://github.com/insightsengineering/teal.modules.general/actions/runs/5401394235/jobs/9811839374?pr=530#step:20:114

Any idea what to do? Should we deleted examples, or should we reduce NOTES check on GitHub Action checks?

gogonzo:
@m7pr Please remove `scda` from dependency and make examples to use iris, mtcars - they would run much faster then. 

Here is the issue (it's a part of our increment goal)
https://github.com/insightsengineering/teal.modules.general/issues/532

m7pr:
sure @gogonzo , will do! thanks

pawelru:
> Any idea what to do? Should we delete examples, or should we reduce NOTES check on GitHub Action checks?

I really don't know. I would like to still keep this note in our block-list if possible, i.e. I agree with @gogonzo to think how to optimize those examples. This is generally not good that example takes long time to execute. FYI: Yet another way of dealing with this is to set up [`_R_CHECK_EXAMPLE_TIMING_THRESHOLD_`](https://rstudio.github.io/r-manuals/r-ints/Tools.html#index-_r_check_example_timing_threshold_) env var

m7pr:
@pawelru @gogonzo what if we setup [_R_CHECK_EXAMPLE_TIMING_THRESHOLD_](https://rstudio.github.io/r-manuals/r-ints/Tools.html#index-_r_check_example_timing_threshold_) to 10 seconds? And for the decoupling, we can have more time to rethink - I would suggest storing minimal SCDA data in teal/data directory and reuse it in all teal.* family packages that use teal. This way we keep good examples, based on clinical/scda data and we got rid of scda

pawelru:
@m7pr I am not that deep in this package so can't say how hard is it to do the optimisation. Please make a call. I trust your assessment

m7pr:
@gogonzo what would you say for bumping up the threshold time for examples in here to 10 seconds?
https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/build-check-install.yaml#L410

And for the decoupling I think we could have a separate discussion as I rather recommend to have scda data stored in teal and reused in mutltiple packages, instead of having examples on iris

chlebowa:
> @gogonzo what would you say for bumping up the threshold time for examples in here to 10 seconds? https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/build-check-install.yaml#L410

I think this is a good temporary solution, the threshold can be lowered once #532 is completed.

> And for the decoupling I think we could have a separate discussion as I rather recommend to have scda data stored in teal and reused in mutltiple packages, instead of having examples on iris

A similar solution was suggested and explicitly rejected.

m7pr:
yeah, I see that, we would need to use `scda::cdisc_data` anyway to pass to `teal::init` if we ever stored scda in teal/data directory. So I guess we go with mtcars/iris.

I made a draft to increase examples time - waiting for @gogonzo for a green light https://github.com/insightsengineering/r.pkg.template/pull/167

pawelru:
> I made a draft to increase examples time - waiting for @gogonzo for a green light https://github.com/insightsengineering/r.pkg.template/pull/167

Please do this for this repo only. I would like to stick to defaults for all the other repos and treat this one as an exception (out of that default value of 5)

m7pr:
ok @pawelru , I made a PR to this repository - maybe you can review and accept https://github.com/insightsengineering/teal.modules.general/pull/533

gogonzo:
I wouldn't keep data in `teal` as we have following dependency order:

- teal.widgets
- teal.code
- teal.data (use scda)
- teal.transform (use scda)
- teal.slice (use scda)
- teal (use scda and all packages above)
- teal.modules.xyz (use teal and teal.transform)

The fact that `teal.transform` uses `scda` is unfortunate and it [will be removed from there soon](https://github.com/insightsengineering/teal.transform/issues/133). In fact, only packages which need artifical clinical data are `teal.modules.clinical`, `osprey`, `teal.osprey`, `goshawk`, `teal.goshawk` and perhaps `teal.modules.hermes`.

----

Optimally, data could be stored in `teal.data` (as name of the package suggests) but `teal.data` in current form is not a dependency of the `teal.modules.xyz`. But in a long term I expect `teal.data` to have ONE reproducible data class which could be used in whole teal application from setting `teal::init(data, ...)` to  `teal.module.xyz::tm_module_srv(data, ...)` - It would make `teal.data` being a dependency of all `teal.xyz` packages. My suggestion is exactly the same as @pawelru - to make a copy of used scda datasets in all packages which needs them.

PR #533, opened 2023-06-29
extend examples timing threshold to 11 seconds

m7pr:


.github/workflows/check.yaml
```diff
@@ -34,6 +34,7 @@ jobs:
     with:
       additional-env-vars: |
         _R_CHECK_CRAN_INCOMING_REMOTE_=false
+        _R_CHECK_EXAMPLE_TIMING_THRESHOLD_: "10"
```
pawelru, line 37: is the format correct? `:` instead of `=` and `"10"` instead of `10`
let's see what R CMD CHECK tells - at the beggining it is reported list of used env vars

m7pr: @pawelru documentation states https://rstudio.github.io/r-manuals/r-ints/Tools.html#index-_r_check_example_timing_threshold_

> _R_CHECK_EXAMPLE_TIMING_THRESHOLD_ [¶](https://rstudio.github.io/r-manuals/r-ints/Tools.html#index-_r_check_example_timing_threshold_)
If timings are being recorded, set the threshold in seconds for reporting long-running examples (either user+system CPU time or elapsed time). Default: "5".

So this "10" should be fine. but `:` might not be proper - let's see the result of the build

m7pr: you were right @pawelru

gogonzo (approved):
Let's change it for two days

PR #541, opened 2023-08-01
propagate argument name change in teal::module

chlebowa:
Part of [this](https://github.com/insightsengineering/teal/issues/769)


gogonzo (approved):
👍

PR #553, opened 2023-08-03
Revert square bracket@main

gogonzo:
reverted replace of square bracket by `get_var`. Problem with `get_var` is as follows.

- `get_var(object = qenv(), object = "varname")` is dispatched on `object` and there are variants for `qenv` and `qenv.error`. Sometimes `qenv()` can be a shiny-validation-error which causes error message like `no get_var method for object error bla bla`.
- `[[` is generic exported from base and we have extra `[[.qenv` and `[[.qenverror` and this function have also other variants created in base - so that it doesn't fail when calling `<simpleError>[["varname"]]`

It means that to fix problem of plot printed to devide we need to use `teal.code::supress_dev()` in specific cases. We don't need to depend on a `get_var`

### DIFF
--- a/R/tm_a_pca.R
+++ b/R/tm_a_pca.R
@@ -919,7 +919,7 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
     })
 
     plot_r <- reactive({
-      teal.code::get_var(output_q(), "g")
+      output_q()[["g"]]
     })
 
     pws <- teal.widgets::plot_with_settings_srv(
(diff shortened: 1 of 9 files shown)


chlebowa (approved):

gogonzo:
> @gogonzo Initially, we converted the [[ call to get_var to prevent the plot from being printed to the IDE. Now, if we're replacing this call, shouldn't we include a call to dev_suppress to maintain the suppression of plot printing?

`dev_supress` should be used only when plot is sent anywhere but plotOutput. We have only one situation in tmg where plot is sent to renderText.

Issue #554, opened 2023-08-04
[Bug]: plots from tm_g_scatterplot and tm_a_regression modules print on IDE


kartikeyakirar:
### What happened?

A bug happened!

The plot appears on the IDE . I believe this issue is associated with the following GitHub pull request: https://github.com/insightsengineering/teal.modules.general/pull/553.

![image](https://github.com/insightsengineering/teal.modules.general/assets/6700955/785f897c-e1be-4fec-88ac-dc5ae2b474c8)

PR #555, opened 2023-08-04
[Bug]: plots from tm_g_scatterplot and tm_a_regression modules print on IDE 554

kartikeyakirar:
this PR fixes https://github.com/insightsengineering/teal.modules.general/issues/554


gogonzo (approved):
👍

Issue #564, opened 2023-08-09
[Feature Request]: Unified Visual Consistency for UI Components


kartikeyakirar:
### Feature description

1 Equalize Outlier Slider Height: Introduce appropriate spacing adjustments between the switch buttons to align them with the outlier slider, ensuring a more uniform and consistent height presentation.

2 Enhance Color Scheme Consistency: Match the colors of the switch buttons with the slider and Reporter button above for a consistent look.

<img width="493" alt="image" src="https://github.com/insightsengineering/teal.modules.general/assets/6700955/1ea591d2-031b-4aa6-b795-5b7536a08dcb">

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

PR #565, opened 2023-08-09
Unified Visual Consistency for UI Components#564

kartikeyakirar:
This fixes https://github.com/insightsengineering/teal.modules.general/issues/564

To ensure a neat and compact arrangement of `switchInput `and `slider`, I've employed a line break. instead of adding new classes. 

image

image

### DIFF
--- a/inst/css/custom.css
+++ b/inst/css/custom.css
@@ -85,3 +85,12 @@
   height:600px;
   width:100%
 }
+
+/* bootstrap rules */
+.bootstrap-switch .bootstrap-switch-primary {
+    background: #337ab7 !important;
+}
+
+.bootstrap-switch.bootstrap-switch-focused {
+    border-color: #2e6da4 !important;
+}
--- a/inst/css/custom.css
+++ b/inst/css/custom.css
@@ -85,12 +85,3 @@
   height:600px;
   width:100%
 }
-
-/* bootstrap rules */
-.bootstrap-switch .bootstrap-switch-primary {
-    background: #337ab7 !important;
-}
-
-.bootstrap-switch.bootstrap-switch-focused {
-    border-color: #2e6da4 !important;
-}
--- a/R/tm_variable_browser.R
+++ b/R/tm_variable_browser.R
@@ -306,6 +306,7 @@ srv_variable_browser <- function(id,
         fluidRow(
           div(
             class = "col-md-4",
+            br(),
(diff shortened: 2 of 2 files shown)


inst/css/custom.css
```diff
@@ -85,3 +85,12 @@
...
 }
+
+/* bootstrap rules */
+.bootstrap-switch .bootstrap-switch-primary {
+    background: #337ab7 !important;
+}
+
+.bootstrap-switch.bootstrap-switch-focused {
+    border-color: #2e6da4 !important;
+}
```
chlebowa, line 96: Are we overwriting theme colors? With hard coded ones?

kartikeyakirar: I did not like it either but to maintain consistency with the other buttons, I modified the CSS property of `'bootstrap-switch_primary'.` While there is a possibility to alter the color using the `'onStatus' `option in `'shinywidget::switchInput',` the available choices are confined, as detailed in the [list provided here](https://getbootstrap.com/docs/4.0/utilities/colors/). Given that 'switchInput' is also employed in 'tm_g_bivariate', I opted for this approach.

As switchInput is also used tm_g_bivariate as welll so I went with this option.

chlebowa: I was under the impression that we should be using consistent colors throughout. Are status colors not enough to achieve this?

gogonzo: I share @chlebowa opinion. Can we instead utilise primary, secondary, etc. everywhere?
I think what you do is against [this](https://github.com/insightsengineering/teal/issues/735) issue?

gogonzo: I'd suggest to explore these (general) issues, making patches here and there:

- https://github.com/insightsengineering/teal/issues/735
- https://github.com/insightsengineering/teal/issues/802
- https://github.com/insightsengineering/teal/issues/762


I think we should first solve general styling of teal apps and then fix wherever style would be incompatible with standard one.

kartikeyakirar: Certainly, I concur. I will address color consistency as a component of these distinct tasks and undo the color modifications from the following pull requests:

Issue #735 - GitHub link: https://github.com/insightsengineering/teal/issues/735
Issue #802 - GitHub link: https://github.com/insightsengineering/teal/issues/802
Issue #762 - GitHub link: https://github.com/insightsengineering/teal/issues/762

Issue #568, opened 2023-08-10
[Bug]:  Inconsistent Styling and Layout for 'Show Warning' and 'Show R Code' Buttons in Module tm_missing_data


kartikeyakirar:
### What happened?

A bug happened!

There is an issue in the module tm_missing_data where the styling and layout for the 'Show Warning' and 'Show R Code' buttons are inconsistent.

<img width="1552" alt="image" src="https://github.com/insightsengineering/teal.modules.general/assets/6700955/6799de37-efe5-4abc-b2b9-168d8a6f3947">

<img width="1552" alt="image" src="https://github.com/insightsengineering/teal.modules.general/assets/6700955/ddf20c25-77f9-4400-95e7-c1dd98f8ea37">

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

PR #573, opened 2023-08-11
Fixing inconsistent Styling and Layout for 'Show Warning' and 'Show R Code' Buttons

kartikeyakirar:
this fixes https://github.com/insightsengineering/teal.modules.general/issues/568

### DIFF
--- a/R/tm_missing_data.R
+++ b/R/tm_missing_data.R
@@ -150,15 +150,17 @@ ui_page_missing_data <- function(id, data, parent_dataname, pre_output = NULL, p
           )
         )
       ),
-      forms = tagList(lapply(
-        datanames,
-        function(x) {
-          conditionalPanel(
-            is_tab_active_js(ns("dataname_tab"), x),
-          div(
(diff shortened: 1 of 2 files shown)


chlebowa:
The implementation is markedly different to that in other modules. Why are the buttons in a conditional panel?

pawelru:
idea: maybe worth to move datasets picker to the column selection area so that we will get rid off one rows for datasets?

kartikeyakirar:
> idea: maybe worth to move datasets picker to the column selection area so that we will get rid off one rows for datasets?

I think this is Good idea. And feasibly, we could remove the 'datasets' subsection under the 'Encoding' label and replace it with a dropdown. However, I'm uncertain about other aspects of the design in relation to this change. Since this falls outside the scope of the current task, I'll create a separate task to initiate a discussion on this matter.

kartikeyakirar:
replaced `tern::summarize_vars` with `tern::analyze_vars` as summarize_vars is deprecated.

chlebowa (approved):

Issue #583, opened 2023-10-03
[Feature Request]: Rewrite modules so they include user's card labels when generating the report


m7pr:
### Feature description

A follow up after https://github.com/insightsengineering/teal.reporter/pull/219

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

PR #584, opened 2023-10-04
198 Include user's card labels when generating the report

kartikeyakirar:
this PR fixes https://github.com/insightsengineering/teal.modules.general/issues/583

 this is follow-up after https://github.com/insightsengineering/teal.reporter/pull/219


.lintr
```diff
@@ -1,5 +1,6 @@
 linters: linters_with_defaults(
   line_length_linter = line_length_linter(120),
   cyclocomp_linter = NULL,
-  object_usage_linter = NULL
+  object_usage_linter = NULL,
+  indentation_linter = NULL
```
m7pr, line 5: @kartikeyakirar is this needed? any way we can revert this indentation NULL?

m7pr (changes requested):

.lintr
kartikeyakirar: without this super linter checks are failing https://github.com/insightsengineering/teal.modules.general/actions/runs/6419052734/job/17428092698 thats I kept this.

m7pr: @kartikeyakirar I rather have you update the code, so that we follow new indentation rules, instead of omitting the check.
Please remove this indentation NULL in .lintr and please edit the code where we need to remove or append lines. Sorry

kartikeyakirar: I had previously attempted this, updating the code in accordance with linter warnings but github auto-action restyle the file.

m7pr: ok, let's leave it with indentation_linter = NULL

m7pr (approved):

m7pr:
Hey @kartikeyakirar looks like this staged deps cyclic error does not appear on this PR anymore. You are good to merge

PR #586, opened 2023-10-16
Replace the `qplot` function with `ggplot` equivalent

vedhav:
Part of fix for [#485](https://github.com/insightsengineering/coredev-tasks/issues/458)

### DIFF
--- a/R/tm_variable_browser.R
+++ b/R/tm_variable_browser.R
@@ -940,7 +940,8 @@ plot_var_summary <- function(var,
 
     if (numeric_as_factor) {
       var <- factor(var, levels = sort(unique(var)))
-      ggplot(NULL, aes(x = var)) + geom_histogram(stat = "count")
+      ggplot(NULL, aes(x = var)) +
+        geom_histogram(stat = "count")
     } else {
       # remove outliers
       if (outlier_definition != 0) {
--- a/R/tm_variable_browser.R
+++ b/R/tm_variable_browser.R
@@ -500,13 +500,11 @@ srv_variable_browser <- function(id,
 
     ### REPORTER
     if (with_reporter) {
-      card_fun <- function(comment, label) {
-        card <- teal::report_card_template(
-          title = "Variable Browser Plot",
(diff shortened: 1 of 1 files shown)


R/tm_variable_browser.R
```diff
@@ -940,7 +940,8 @@ plot_var_summary <- function(var,
 
     if (numeric_as_factor) {
       var <- factor(var, levels = sort(unique(var)))
```
chlebowa, line 942: I don't think the `sort(unique(` is necessary here.

chlebowa (approved):

R/tm_variable_browser.R
chlebowa: Levels will always be unique and levels being sorted is the default behavior. One runs `levels = unique(x)` to prevent them from being sorted.
```
x <- letters[4:1]
factor(x)
factor(x, levels = unique(x))
```

chlebowa: It's not that big a deal though.

vedhav: Oh! I prefer just using unique in that case. Thanks for suggesting.

vedhav: But I think it's best to keep it this way:
```r
> var <- c(1, 3, 2, 1, 3, 1, 2, 1, 2)
> var <- factor(var, levels = sort(unique(var)))
> var
[1] 1 3 2 1 3 1 2 1 2
Levels: 1 2 3
> var <- c(1, 3, 2, 1, 3, 1, 2, 1, 2)
> var <- factor(var, levels = unique(var))
> var
[1] 1 3 2 1 3 1 2 1 2
Levels: 1 3 2
```

chlebowa: Ah, but the sorting may makes sense in this case, I think. You represent a numeric as a factor and you draw a histogram.

vedhav: Okay, this is perfect! I agree that this is better.

PR #585, opened 2023-10-13
812 landing_popup

m7pr:
Partner of https://github.com/insightsengineering/teal/pull/934/

PR #589, opened 2023-10-23
812 landing popup@main

m7pr:
Accidentally closed https://github.com/insightsengineering/teal.modules.general/pull/585 which can not be opened. So moving again in here


R/tm_landing_popup.R
```diff
@@ -0,0 +1,82 @@
...
+    checkmate::assert_multi_class(
+      content,
+      classes = c("character", "shiny.tag", "shiny.tag.list", "html"), null.ok = TRUE
+    )
+    checkmate::assert_multi_class(buttons, classes = c("shiny.tag", "shiny.tag.list"))
+
+    logger::log_info("Initializing tm_landing_popup")
+
+    module(
+      label = label,
```
m7pr, line 67: passing label on purpose. even though there's not gonna be any tab for this module, one maybe would like to extract this module by label.

R/tm_landing_popup.R
```diff
@@ -0,0 +1,82 @@
+#' Landing Popup Module
+#'
+#' @description This module creates a landing welcome popup for `teal` applications.
+#'
+#' @details If you use this module in `teal::init(modules = )`, it will not be wrapped in a tab in `teal` application.
+#' See `teal::module(type = 'teal_module_landing')`.
+#'
+#' @param label `character(1)` the label of the module.
+#' @param title `character(1)` the text to be displayed as a title of the popup.
+#' @param content `character(1)` the content of the popup. Passed to `...` of `shiny::modalDialog`. Can be a `character`
```
chlebowa, line 10:
```suggestion
#' @param content the content of the popup. Passed to `...` of `shiny::modalDialog`. Can be a `character`
```

R/tm_landing_popup.R
```diff
@@ -0,0 +1,82 @@
...
+#' or a text input control (like `textInput`) or a list of `shiny` tags. See examples.
+#' @param buttons `shiny` tag or a list of tags (`tagList`). Typically a `modalButton` or `actionButton`. See examples.
```
chlebowa, line 12:
```suggestion
#' @param buttons `shiny.tag` or a list of tags (`tagList`). Typically a `modalButton` or `actionButton`. See examples.
```

R/tm_landing_popup.R
```diff
@@ -0,0 +1,82 @@
...
+#' if (interactive()) {
+#'   shinyApp(app1$ui, app1$server)
+#' }
+#'
+#' app2 <- teal::init(
+#'   data = teal.data::dataset("iris", iris),
+#'   modules = teal::modules(
+#'     teal.modules.general::tm_landing_popup(
+#'       title = "Welcome",
+#'       content = div(tags$b("A place for the welcome message or a disclaimer statement.", style = "color: red;")),
```
chlebowa, line 36: `div` is redundant, I think.

R/tm_landing_popup.R
```diff
@@ -0,0 +1,82 @@
...
+#'       buttons = tagList(
+#'         modalButton("Proceed"),
+#'         actionButton("close", "Read more", onclick = "window.open('http://google.com', '_blank')")
```
chlebowa, line 39: Perhaps create an example with a "Reject" button that will close the app? Navigate to a new tab in 5 seconds?

R/tm_landing_popup.R
```diff
@@ -0,0 +1,82 @@
...
+           title = NULL,
+           content = NULL,
+           buttons = modalButton("Accept")) {
+    checkmate::assert_string(label)
+    checkmate::assert_string(title, null.ok = TRUE)
+    checkmate::assert_multi_class(
+      content,
+      classes = c("character", "shiny.tag", "shiny.tag.list", "html"), null.ok = TRUE
+    )
+    checkmate::assert_multi_class(buttons, classes = c("shiny.tag", "shiny.tag.list"))
```
chlebowa, line 62: There is a conundrum. The modal can be closed either by a `modalButton` or a regular `actionButton` as long as there is an observer that will call `removeModal`. We can easily assert the former but not the latter. And I think some assertion should be made, otherwise the app dev can create a popup that cannot be dismissed.

R/tm_landing_popup.R
chlebowa: I think we will rather move to extracting solely by class but I see no harm in passing the label.

R/tm_landing_popup.R
m7pr: https://github.com/insightsengineering/teal/pull/934/commits/8f111e847bbf2c7bd3113e44e1b76062c58f2985

R/tm_landing_popup.R
m7pr: Yup, just extended. Good idea
https://github.com/insightsengineering/teal/pull/934/commits/4eb1a39995b910efb69aac5a853479fb92e55ace

R/tm_landing_popup.R
m7pr: yeah, if there is no modalButton then this might not be possible to close the modal, but someone can invent an actionButton that does something extra + be closing the modal :P
Unsure if we should force the usage of modalButton at least once

R/tm_landing_popup.R
m7pr: https://github.com/insightsengineering/teal/pull/934/commits/c6a02031f8b01f66750039cf65a0803713418760

R/tm_landing_popup.R
chlebowa: I mean it makes sense to make sure that the modal can be closed but it's no trivial matter.

m7pr: > otherwise the app dev can create a popup that cannot be dismissed.

Yeah, but (s)he should notice that and can figure this out on his/her own.

m7pr: I don't think you will ever be able to address if this can be closed : P

R/tm_landing_popup.R
m7pr: totally. but still the `modules()/teal_modules` object can be used in `modules_labels` where label can be useful (somehow). this is just an abstract idea

chlebowa: Where? :thinking:

R/tm_landing_popup.R
chlebowa: All right, let's trust them.

m7pr:
Closing in favour of https://github.com/insightsengineering/teal/pull/934

Issue #619, opened 2023-12-18
[Feature Request]: update modules to handle default `datanames` when datanames = "all"


kartikeyakirar:
### Feature description

Many modules require datasets for encodings. If` datanames` are not set, the default from the environment is used. However, some modules such as `tm_data_table,` `tm_variable_browser`,and `tm_missing_data` take everything, which causes errors or app failures. 

simpler fix would be the rendering of data that are not inheriting from the `data.frame` should be skipped. 

Issues for setting default `datanames` and handling filterpanel for anything other than `data.frame` and `MAE.`
https://github.com/insightsengineering/teal/issues/999 
https://github.com/insightsengineering/teal.slice/issues/493

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

PR #621, opened 2023-12-19
619 update modules to handle default datanames when datanames = "all"

kartikeyakirar:
this fixes https://github.com/insightsengineering/teal.modules.general/issues/619

Here, I am filtering `datanames `to only allow` data.frame` class and skipping other classes. Should we notify the user about datanames which are skipped ?

### DIFF
--- a/R/tm_data_table.R
+++ b/R/tm_data_table.R
@@ -1,6 +1,7 @@
 #' Data Table Viewer Teal Module
 #'
 #' A data table viewer shows the data using a paginated table.
+#' specifically designed for use with `data.frames`.
 #' @md
 #'
 #' @inheritParams teal::module
@@ -12,6 +13,8 @@
 #' @param datasets_selected (`character`) A vector of datasets which should be
 #'   shown and in what order. Names in the vector have to correspond with datasets names.
 #'   If vector of length zero (default) then all datasets are shown.
+#'   Note: Only datasets of the `data.frame` class are compatible;
+#'   using other types will cause an error.
 #' @param dt_args (named `list`) Additional arguments to be passed to `DT::datatable`
 #'   (must not include `data` or `options`).
 #' @param dt_options (named `list`) The `options` argument to `DT::datatable`. By default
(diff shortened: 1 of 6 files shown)


gogonzo (commented):
Please add in the modules' documentation that they support `data.frames` only. In `tm_missing_data` and `tm_data_table` there is `datasets_selected` argument - I think there should be mentioned that it supports data.frames only.

R/tm_variable_browser.R
```diff
@@ -17,6 +18,7 @@
 #' @param datasets_selected (`character`) A vector of datasets which should be
 #'   shown and in what order. Names in the vector have to correspond with datasets names.
 #'   If vector of length zero (default) then all datasets are shown.
+#'   Note: Only datasets of the `data.frame` class are compatible; using other types will cause an error.
```
gogonzo, line 21:
```suggestion
#'   Note: Only datasets of the `data.frame` class are compatible; other data tupés will be skipped.
```

R/tm_data_table.R
```diff
@@ -12,6 +13,8 @@
 #' @param datasets_selected (`character`) A vector of datasets which should be
 #'   shown and in what order. Names in the vector have to correspond with datasets names.
 #'   If vector of length zero (default) then all datasets are shown.
+#'   Note: Only datasets of the `data.frame` class are compatible;
+#'   using other types will cause an error.
```
gogonzo, line 17:
```suggestion
#'   Note: Only datasets of the `data.frame` class are compatible; other data tupés will be skipped.

```

Not true. Other types will be skipped as there is a Filter call

gogonzo (approved):
Small changes but important i think

R/tm_data_table.R
kartikeyakirar: for `dataset_selected `if its not `data.frame` then it will fail assertion https://github.com/insightsengineering/teal.modules.general/blob/975d06400514b647d6f1f7c429b39611a5073b61/R/tm_data_table.R#L151-L155. other types are only skipped when set as` datanames `not `dataset_selected`

R/tm_variable_browser.R
kartikeyakirar: Other datatypes for `dataset_selected` will fail assertion. Do we also want to skip other data types for the `dataset_selected `parameter?

R/tm_data_table.R
gogonzo: Ok, fair enough

gogonzo: Feel free to merge

PR #626, opened 2024-01-24
options for strict tests; few enhancements

pawelru:
- part of https://github.com/insightsengineering/coredev-tasks/issues/498
- part of https://github.com/insightsengineering/coredev-tasks/issues/478
  - please read this for more info about the implementation: https://github.com/insightsengineering/coredev-tasks/issues/478#issuecomment-1909912778
- removed `pkg::` from examples for packages in Depends
- reoredered rd tags to have examples at the bottom
- fix roxygen warning about documenting package level Rd file
- removed `# nolint` from code-docs (e.g. examples or knitr chunks) (this is actually rendered!). I needed to put a higher level exclusion in the top `.lintr` file. This is of course more powerful than single line (or code-bloc) exclusions
- reviewed `dontrun{}`
- removed `teal.modules.general::` from examples and vignettes as this is actually redundant here

Please review the changes carefully and let me know if there is something you don't like.


.lintr
```diff
@@ -4,3 +4,35 @@ linters: linters_with_defaults(
...
+    "vignettes/using-response-plot.Rmd" = list(
+      object_name_linter = Inf
+    ),
+    "vignettes/using-scatterplot-matrix.Rmd" = list(
+      object_name_linter = Inf
+    ),
+    "vignettes/using-scatterplot.Rmd" = list(
+      object_name_linter = Inf
+    )
+  )
```
vedhav, line 38: During the pkg cleanup we agreed that a common regex for `object_name_linter` should be the way to go.
So, we went with:
```
object_name_linter = object_name_linter(styles = c("snake_case", "symbols"), regexes = c(ANL = "^ANL_?[0-9A-Z_]*$", ADaM = "^r?AD[A-Z]{2,3}_?[0-9]*$"))
```
This solves our object name lint issue. Can we remove these file-level exclusions?

vedhav (approved):
Other than the `.lintr` config the changes LGTM! We can merge it after it is resolved.

.lintr
pawelru: Yes I think so. Let me do that

vedhav: Thanks! Also thanks for fixing warning with `color_setting`

PR #628, opened 2024-02-01
register teal.logger handlers

pawelru:
close https://github.com/insightsengineering/coredev-tasks/issues/502
test with https://github.com/insightsengineering/teal.logger/pull/73

As a bonus, we can safely move `logger` to suggests (it's still used in tests)


R/tm_a_pca.R
```diff
@@ -105,7 +105,7 @@ tm_a_pca <- function(label = "Principal Component Analysis",
                      size = c(2, 1, 8),
                      pre_output = NULL,
                      post_output = NULL) {
-  logger::log_info("Initializing tm_a_pca")
+  message("Initializing tm_a_pca")
```
gogonzo, line 108: image

`messages` are logged as is. Looks like changes in `teal.logger` do not have any effect

gogonzo (approved):
Tusind tak!

DESCRIPTION
```diff
@@ -45,7 +44,7 @@ Imports:
     stringr (>= 1.4.1),
     teal.code (>= 0.5.0),
     teal.data (>= 0.5.0),
-    teal.logger (>= 0.1.1),
+    teal.logger (>= 0.1.3.9011),
```
pawelru, line 47:
```suggestion
    teal.logger (>= 0.1.3.9013),
```

R/tm_a_pca.R
pawelru: this was fixed in teal.logger

PR #636, opened 2024-02-14
635 remove test theme

### LINKED ISSUE #635: [Question]: is `ggplot2::theme_test()`  a real `theme`?
We do allow to pass various `{ggplot2}` themes in our encoding panel, for example in `tm_missing_data`
image

Is this theme really o ggplot2 theme that is meaningful?
`?ggplot2::theme_test` states

> A theme for visual unit tests

chlebowa:
Closes #635 

Added global variable, `ggplot_themes`, that lists all `ggplot2` themes in a character vector.
`ggplot_themes` is used in all calls to `selectInput` that present a choice of themes.

Removed `"test"` from default arguments that specify initial theme.

Trimmed down formal argument descriptions.

Also, restyled the package documentation file to the currently recommended way (`"_PACKAGE"` replaceas `@docType package`).

Additionally, changed `# nolint` tags to state specific linters that are omitted and removed superfluous `#nolint` tags.


R/zzz.R
```diff
@@ -1,3 +1,5 @@
 .onLoad <- function(libname, pkgname) { # nolint
   teal.logger::register_logger(namespace = "teal.modules.general")
 }
+
+ggplot_themes <- c("gray", "bw", "linedraw", "light", "dark", "minimal", "classic", "void")
```
m7pr, line 5: Hey, this PR is excellent and removes a lot of repetitions in the code. I like the idea of having a variable storing all themes that we reuse. The only, tiny, hesitance about this one is the possibility that a variable called `ggplot_themes` already exists in someone's session when tmg is loaded. Then there is a conflict.

What if (pick one, my preference is the `options`)
- we change the name from `ggplot_themes` to `teal_ggplot_themes`
- or we set an option `options(teal_ggplot_themes = c("gray", "bw", "linedraw", "light", "dark", "minimal", "classic", "void"))` and we use `getOpion('teal_ggplot_themes')` in `selectInput`
- we create a hidden environment called `.teal_transform_options <- env()` where we assign `ggplot_themes` and then we use `get('ggplot_themes', env = '.teal_transform_options')` in `selectInput`

m7pr:
This is great, left just one comment, that I think we can over-come quickly

R/zzz.R
chlebowa: `teal_ggplot_themes` would indicate that these themes are somehow connected to or defined in `teal`, which is not the case: it is merely an enumeration of themes provided by `ggplot2`.

Functions will scope their enclosing environment first before scoping the global environment. See here:
```
library(teal)
devtools::load_all()

ggplot_themes <- c("gray", "void")

data <- teal_data()
data <- within(data, {
  library(nestcolor)
  ADSL <- teal.modules.general::rADSL
})
datanames <- c("ADSL")
datanames(data) <- datanames
join_keys(data) <- default_cdisc_join_keys[datanames]
app <- teal::init(
  data = data,
  modules = teal::modules(
    teal.modules.general::tm_a_pca(
      "PCA",
      dat = teal.transform::data_extract_spec(
        dataname = "ADSL",
        select = teal.transform::select_spec(
          choices = teal.transform::variable_choices(
            data = data[["ADSL"]], c("BMRKR1", "AGE", "EOSDY")
          ),
          selected = c("BMRKR1", "AGE"),
          multiple = TRUE
        ),
        filter = NULL
      ),
      ggplot2_args = teal.widgets::ggplot2_args(
        labs = list(subtitle = "Plot generated by PCA Module")
      )
    )
  )
)
runApp(app, launch.browser = TRUE)
```

m7pr: My bad, I though `ggplot_themes` will be assigned to `.GlobalEnv()` when this package is loaded. It's gonna be just in tmg namespace, so it will not pollute global environment. Sorry for the confusion. Alrighty so I think this is good to go

m7pr (approved):
👍

averissimo:
The linter changes on this PR are not effective as they need to be in the `# nolint: <linter_rule>.`

The colon `:` and dot `.` at the end are fundamental, otherwise, it behaves as `# nolint` and ignores the following text.

chlebowa:
You're absolutely right!
https://github.com/insightsengineering/teal.modules.general/pull/638

averissimo:
You were fast, I was going to pick this up as I don't like saying this kind of thing and then have someone else do the boring job.

chlebowa:
Oh no, I must fix my mistakes myself, thank you for pointing it out :slightly_smiling_face:

Issue #624, opened 2024-01-17
Pre-release activities


donyunardi:
## Package level checks
* https://github.com/insightsengineering/teal.modules.general/pull/639
* Review and update:
    * README.md (check example code)
    * NEWS.md
* Run urlchecker::url_check() to identify broken links and fix
* Run R CMD check --as-cran make sure everything pass
* Package `Title` is not duplicated in Package `Description` in DESCRIPTION file (e.g. this happens in teal.slice currently)
* All package names in `Title` and `Description` fields of DESCRIPTION file are quoted with `'`
* You have checked the Package Release Template https://github.com/insightsengineering/teal.reporter/pull/205/files
* Make sure all `teal.*` mentions are lower-cased and quoted
* Make Sure inst/WORDLIST is minimalized
* Make sure each link to our documentation hosted with pkgdown on github pages do not have `/main/` in the address but it has `/latest/` instead, so we always expose the documentation of the latest release and not what's currently on main branch but not yet released,
* Remove exception in .lintr: indentation_linter = NULL
* Remove @noMd (in favor of Roxygen: list(markdown = TRUE) in DESCRIPTION)

## Vignettes level checks
* https://github.com/insightsengineering/teal.modules.general/issues/640
* Sanity check of all vignettes, make sure there is no typo, no wrong format, etc.

## Non-exported functions level checks
* https://github.com/insightsengineering/teal.modules.general/issues/641
* Make sure non-exported functions do not have examples

## Exported functions level checks
* Review functions:
    * @example tag, make sure it runs, fix if otherwise
    * Make sure functions has @return tag to document the return value
    * no \dontrun tag, replace with if(interactive()) if needed
    *  Standard order of roxygen2 tags
    * `@title ➡️ @doctype ➡️ @description ➡️ @details ➡️ @rdname ➡️` 
    * `➡️ @inheritParams ➡️ @params ➡️ @return ➡️ @seealso ➡️ @references ➡️`
    * `➡️ @examples ➡️ @export ➡️ @keywords  ➡️  @noRd`
* Make sure there are no `:::` in examples
    * if you need to retain the example that uses `:::`, use `getFromNamespace()` function.
* Remove return wrapper if it is the last expression (per NEST guidelines)

Please self-assign by your name next to the module and link to the PR:
- [x] tm_g_response.R @averissimo #666
- [x] tm_file_viewer.R @averissimo #665
- [x] tm_t_crosstable.R @averissimo #664
[...]

PR #646, opened 2024-02-19
pre-release `tm_a_pca` document update

kartikeyakirar:

~~blocked by https://github.com/insightsengineering/teal.modules.general/pull/651~~


R/tm_a_pca.R
```diff
@@ -1,54 +1,61 @@
...
-#' @param size optional, (`numeric`) If scalar then the plot point sizes will have a fixed size.
-#'   If a slider should be presented to adjust the plot point sizes dynamically then it can be a
-#'   vector of length three with `c(value, min, max)`.
-#' @param font_size optional, (`numeric`) font size control for title, x-axis label, y-axis label and legend.
-#'   If scalar then the font size will have a fixed size. If a slider should be presented to adjust the plot
-#'   point sizes dynamically then it can be a vector of length three with `c(value, min, max)`.
-#'
+#'   specifying columns used to compute PCA.
+#' @param alpha (`numeric`, optional) value or a vector of length three in the form `c(value, min, max)`.
+#' - If provided as a scalar, it sets a fixed opacity for plot points.
```
m7pr, line 13: @kartikeyakirar I don't think we have `scalars` in R. Maybe differentiate by `length`.

> If provided o `length==1`, it sets a fixed...
> If provided as a vector of `length>1`, it allows dynamic...

R/tm_a_pca.R
```diff
@@ -1,54 +1,61 @@
...
+#' - If provided as a vector, it enables dynamic adjustment of font size via a slider in the UI.
 #' @templateVar ggnames "Elbow plot", "Circle plot", "Biplot", "Eigenvector plot"
 #' @template ggplot2_args_multi
 #'
 #' @export
 #'
 #' @examples
 #'
-#' # ADSL example
+# ADSL example
```
m7pr, line 29: I think it was fine. You deleted `#'`
```suggestion
#' # ADSL example
```

R/tm_a_pca.R
```diff
@@ -1,54 +1,61 @@
...
+#' library(teal.widgets)
```
m7pr, line 30: Is `teal.widgets` enough?
Just checking. `teal` and `teal.transform` are in `Depends` right so will be already on `search` path?

R/tm_a_pca.R
m7pr: Yeah, I know `checkmate::assert_scalar` exists, but I don't think we have scalars in R. Unsure if `scalar` is common for users who never seen `checkmate`. I would strive from using it in terms of R documenation if possible and would substitute with something else, just in case

m7pr:
@kartikeyakirar can you wait until thiss https://github.com/insightsengineering/teal.modules.general/pull/651 is done and pull changes to your branch?

R/tm_a_pca.R
m7pr: image

kartikeyakirar: done! [057ae53](https://github.com/insightsengineering/teal.modules.general/pull/646/commits/057ae536febdf3ce45c4f7333e080fb3dc844c00)

R/tm_a_pca.R
```diff
@@ -1,32 +1,37 @@
...
+#' - If vector of `length==1` then the font size will have a fixed size.
+#' - If vector of `length>1`, it enables dynamic adjustment of font size via a slider in the UI.
 #' @templateVar ggnames "Elbow plot", "Circle plot", "Biplot", "Eigenvector plot"
 #' @template ggplot2_args_multi
 #'
 #' @examples
-#' # general data example
 #' library(teal.widgets)
 #'
+#' # General data example
```
m7pr, line 28: I think it's fine if this starts from lower_case `g`
```suggestion
#' # general data example
```

R/tm_a_pca.R
```diff
@@ -171,7 +174,8 @@ tm_a_pca <- function(label = "Principal Component Analysis",
   )
 }
 
-
+# UI function for the PCA module
+# This function defines the UI elements for the PCA analysis, including plot settings and data selection.
```
m7pr, line 178: I'm not sure this is needed

R/tm_a_pca.R
```diff
@@ -282,6 +286,8 @@ ui_a_pca <- function(id, ...) {
   )
 }
 
+# Server function for the PCA module
+# This function handles the server-side logic for PCA analysis, including data processing and plot generation.
```
m7pr, line 290: Unsure if this is needed.

R/tm_a_pca.R
```diff
@@ -1,32 +1,37 @@
...
-#'   length three with `c(value, min, max)`.
-#' @param size optional, (`numeric`) If scalar then the plot point sizes will have a fixed size.
-#'   If a slider should be presented to adjust the plot point sizes dynamically then it can be a
-#'   vector of length three with `c(value, min, max)`.
-#' @param font_size optional, (`numeric`) font size control for title, x-axis label, y-axis label and legend.
-#'   If scalar then the font size will have a fixed size. If a slider should be presented to adjust the plot
-#'   point sizes dynamically then it can be a vector of length three with `c(value, min, max)`.
-#'
+#'   specifying columns used to compute PCA.
+#' @param alpha (`numeric`, optional) value or a vector of length three in the form `c(value, min, max)`.
```
averissimo, line 12: @insightsengineering/nest-core-dev We should align on how to write this "optional" indicator.

It seems we have quite a few usages on all of the teal packages _(even after all releases)_. Here are some that I've gathered from `{teal}`, `{teal.slice}` and `{teal.transform}`

1. ``@param optional (`character`) ...``
1. ``@param (optional `character`) ...``
1. ``@param (`character`, optional) ...``
1. ``@param (`character`) optional ...``

I've been using `2.` _(I think I saw that on the first `pre-release`)_ as it gives some flexibility when the parameter accepts multiple data types, but I think any of them is fine as long as we decide on a single one.

chlebowa: If the type itself is not to be part of the following description, then (4) seems natural.

kartikeyakirar: I am okay with either of them
> @param optional (`character`) ...
 it could be mistaken as part of the parameter 

> @param (optional `character`) ...
> @param (`character`, optional) ...

my pick is adding "optional" before or after the data within parentheses as users often look for the type after/before its optional status.  its easy for the scan.

> `@param (character) optional ...`
it could be mistaken for part of description.

m7pr: I also prefer 4 and 1, over 2 and 3, so I'm fine with 4. @kartikeyakirar would you unify on your PRs in this package, and maybe you can create issues to get this covered in other packages? However I think we should start a discussion in a separate issue, so other team members are aware (in case they dont get notifications from review threads).

averissimo: Cool, let's do `4.` on this package and revisit the other on the near future

R/tm_a_pca.R
kartikeyakirar: I don't see any problem with including additional details as this information is solely meant for developers who navigate the code rather than for formal documentation. but , I will remove line 178. I have kept these detailed in this module only, rest only have titles.

also I think we should also add` @noRd` and `@keyowrd internal `

R/tm_a_pca.R
kartikeyakirar: I will  check which packages needs the changes and log issue for this. I will make changes for this in tmg as separate task  as `shared_param `and `ggplot2_args_multi ` also needs to be edited

averissimo: Or do you think this should warrant further discussion on its own issue? @m7pr @kartikeyakirar

averissimo: We could do `4.` for now _(for consistency sake and merge the PRs)_ and revert early next week with find+replace if we come to a different conclusion

kartikeyakirar: For `tmg` https://github.com/insightsengineering/teal.modules.general/issues/674

For the rest of the package https://github.com/insightsengineering/nestdevs-tasks/issues/52

kartikeyakirar: > Or do you think this should warrant further discussion on its own issue? @m7pr @kartikeyakirar

if needed we could continue this here https://github.com/insightsengineering/nestdevs-tasks/issues/52

R/tm_a_pca.R
m7pr: I think the text is unneeded as you can figure out the purpose of the function just based on the name.

m7pr: I see you already have `noRd` and `keywords internal` in other PRs, so let's have it, if you already made some work on it

R/tm_a_pca.R
m7pr: I see you already have `noRd` and `keywords internal` in other PRs, so let's have it, if you already made some work on it

kartikeyakirar:
removed `@keywords` tag

PR #656, opened 2024-02-21
pre-release `tm_g_association` document update

kartikeyakirar:


R/tm_g_association.R
```diff
@@ -1,27 +1,31 @@
...
+#'
+#'
+#' @note For more examples, please see the vignette "Using association plot" via
+#' `vignette("using-association-plot", package = "teal.modules.general")`.
 #'
 #' @inheritParams teal::module
 #' @inheritParams shared_params
 #' @param ref (`data_extract_spec` or `list` of multiple `data_extract_spec`)
-#'   reference variable, must set `multiple = FALSE`.
+#' Reference variable, must set `multiple = FALSE`.
```
vedhav, line 14:
```suggestion
#' Reference variable. It can only have single selection options in the `select_spec` by setting `multiple = FALSE`.
```

The part `must set 'multiple = FALSE'` was hard to understand for me. A little more context would be helpful. What do you think about something like this?

kartikeyakirar: this would be more clear .WDYT

`#' Reference variable, must set `multiple = FALSE` to ensure each `data_extract_spec` specifies only one variable, preventing multiple selections within a single specification.`

vedhav: `multiple` is the argument of `select_spec()` and not `data_extract_spec` so not mentioning about `select_spec` might confuse people.

kartikeyakirar: > #' Reference variable, must accepts a `data_extract_spec` with `select_spec(multiple = FALSE) `to ensure single selection option.

 how about this ?

vedhav: I like this one 👍🏽 It's more actionable.

vedhav (approved):
👌🏽

PR #651, opened 2024-02-20
29 add non-cdisc examples to the modules

chlebowa:
Closes #29 

Added general data examples to modules.

PR #660, opened 2024-02-21
pre-release `tm_g_bivariate` document update

kartikeyakirar:

blocked by https://github.com/insightsengineering/teal.modules.general/pull/651

### DIFF
--- a/R/tm_g_bivariate.R
+++ b/R/tm_g_bivariate.R
@@ -1,50 +1,55 @@
-#' Univariate and bivariate visualizations
-#' @md
+#' Univariate and bivariate visualizations module
 #'
-#' @inheritParams teal::module
-#' @inheritParams shared_params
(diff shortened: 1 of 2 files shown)


R/tm_g_bivariate.R
```diff
@@ -104,6 +109,10 @@
 #' if (interactive()) {
 #'   shinyApp(app$ui, app$server)
 #' }
+#'
+#'
+#' @export
+#'
```
pawelru, line 115: These empty spaces will be actually rendered out. Suggest to remove and keep the example code minimal.

chlebowa: I mean the ones below `@export`. If you're referring to double empty lines above `@export`, I agree.

pawelru: yep, the ones above - I misclicked and highlighted more that I should :)

m7pr (approved):
👍

PR #661, opened 2024-02-21
pre-release `tm_g_scatterplotmatrix` document update

kartikeyakirar:


R/tm_g_scatterplotmatrix.R
```diff
@@ -524,7 +532,8 @@ get_scatterplotmatrix_stats <- function(x, y,
       if (.f_args$method == "spearman") {
         return("rho:-")
       }
+    } else {
+      return("-")
     }
```
averissimo, line 537: Can you revert this?

The previous implementation returned `"-"` when both there is no method or it has a name that is different than the options.

Either that or replace the contents of `else` clause with: 

```r
  # ...
  } else {
    checkmate::assert_list(.f_args)
    switch(
      ifelse(is.null(._f_args$method), "", ._f_args$method),
      "pearson" = "cor:-",
      "kendall" = "tau:-",
      "spearman" = "rho:-",
      "-"
    )
  }
}
```

R/tm_g_scatterplotmatrix.R
```diff
@@ -461,26 +463,29 @@ srv_g_scatterplotmatrix <- function(id, data, reporter, filter_panel_api, variab
...
+#' ```
+#' @param x (`numeric`)
+#' @param y (`numeric`)
+#' @param .f (`function`) function that accepts x and y as formula input `~ x + y`.
+#' Default `stats::cor.test`.
+#' @param .f_args (`list`) of arguments to be passed to `.f`.
+#' @param round_stat (`integer`)
+#' @param round_pval (`integer`)
+#'
+#' @return Character with stats. For `stats::cor.test` correlation coefficient and p-value.
```
averissimo, line 488: (optional suggestion)

Given that the close relationship between the functions I suggest linking it.

```suggestion
#' @return Character with stats. For [stats::cor.test()] correlation coefficient and p-value.
```

R/tm_g_scatterplotmatrix.R
```diff
@@ -461,26 +463,29 @@ srv_g_scatterplotmatrix <- function(id, data, reporter, filter_panel_api, variab
...
-#' @param .f \code{function}, function that accepts x and y as formula input \code{~ x + y}.
-#' Default \code{stats::cor.test}
-#' @param .f_args \code{list} of arguments to be passed to \code{.f}
-#' @param round_stat \code{integer}
-#' @param round_pval \code{integer}
-#' @details presently we need to use a formula input for \code{stats::cor.test} because
-#' \code{na.fail} only gets evaluated when a formula is passed (see below).
-#' \preformatted{
+#'
+#' Uses `stats::cor.test` per default for all numerical input variables and converts results
```
averissimo, line 467: (optional suggestion)

Given that the close relationship between the functions I suggest linking it.

```suggestion
#' Uses [stats::cor.test()] per default for all numerical input variables and converts results
```

R/tm_g_scatterplotmatrix.R
```diff
@@ -461,26 +463,29 @@ srv_g_scatterplotmatrix <- function(id, data, reporter, filter_panel_api, variab
...
+#' to character vector.
+#' Could be extended if different stats for different variable types are needed.
+#' Meant to be called from `lattice::panel.text`.
```
averissimo, line 470: (optional suggestion)

Given that the close relationship between the functions I suggest linking it.

```suggestion
#' Meant to be called from [lattice::panel.text()].
```

R/tm_g_scatterplotmatrix.R
```diff
@@ -461,26 +463,29 @@ srv_g_scatterplotmatrix <- function(id, data, reporter, filter_panel_api, variab
...
 #' x = c(1,3,5,7,NA)
 #' y = c(3,6,7,8,1)
 #' stats::cor.test(x, y, na.action = "na.fail")
 #' stats::cor.test(~ x + y,  na.action = "na.fail")
-#' }
-#' @return \code{character} with stats. For \code{stats::cor.test} correlation coefficient and p-value.
-#' @export
+#' ```
+#' @param x (`numeric`)
+#' @param y (`numeric`)
```
averissimo, line 481: Should there be a description here?

(suggestion copies it from `stats::corr` and adds the data type)

```suggestion
#' @param x,y (`numeric`) vectors of data values. `x` and `y` must have the same length.
```

R/tm_g_scatterplotmatrix.R
```diff
@@ -461,26 +463,29 @@ srv_g_scatterplotmatrix <- function(id, data, reporter, filter_panel_api, variab
...
+#' @param .f (`function`) function that accepts x and y as formula input `~ x + y`.
+#' Default `stats::cor.test`.
+#' @param .f_args (`list`) of arguments to be passed to `.f`.
+#' @param round_stat (`integer`)
+#' @param round_pval (`integer`)
```
averissimo, line 486:
```suggestion
#' @param round_stat (`integer(1)`) optional. Number of decimal places to use when rounding the estimate.
#' @param round_pval (`integer(1)`) optional. Number of decimal places to use when rounding the p-value.
```

averissimo (changes requested):
I believe the code change should be reverted.

Added a few comments on documentation that I think either need to be changed, or would improve it (marked with _optional_).

R/tm_g_scatterplotmatrix.R
kartikeyakirar: 😱 I think I am lost in {}.

kartikeyakirar: [b9db727](https://github.com/insightsengineering/teal.modules.general/pull/661/commits/b9db72789d12ec0d57bd3b8ee81684c863a2ec8d)

averissimo (approved):
Let's go!

PR #664, opened 2024-02-21
pre-release `tm_t_crosstable` document update

averissimo:
part of #624

~blocked by #651~


R/tm_t_crosstable.R
```diff
@@ -167,6 +174,7 @@ tm_t_crosstable <- function(label = "Cross Table",
   )
 }
 
+# UI function for the cross-table module
```
m7pr, line 177: @averissimo in other PRs @kartikeyakirar put `@noRd` and `@keywords internal` for those `ui_` and `server_` functions. Can we have them here as well, so it's unified... sorry?

m7pr:
@averissimo I think you are good merging this one.

PR #669, opened 2024-02-22
pre-release `tm_g_distribution` document update

kartikeyakirar:


R/tm_g_distribution.R
```diff
@@ -172,6 +170,9 @@ tm_g_distribution <- function(label = "Distribution Module",
   )
 }
 
+# UI function for the distribution module.
+#' @noRd
+#' @keywords internal
```
vedhav, line 175: Let's resolve the decision on how we should be using the `@noRd` and `@keywords internal` tags.
The previous 3 PRs that I reviewed did not add any tags.

m7pr: yes, this should be removed

R/tm_g_distribution.R
```diff
@@ -1,33 +1,33 @@
...
-#'   vector of length three with `c(value, min, max)`.
-#'   Defaults to `c(30L, 1L, 100L)`.
+#' Categorical variable used to split the distribution analysis.
+#' @param group_var (`data_extract_spec` or `list` of multiple `data_extract_spec`)
+#' Variable used for faceting plot into multiple panels.
+#' @param freq (`logical`, optional) Whether to display frequency (`TRUE`) or density (`FALSE`).
+#' Defaults to density (`FALSE`).
+#' @param bins (`integer(1)` or `integer(3)`, optional)  Determines the number of bins for the histogram.
+#' - If vector of `length == 1` then the histogram bins will have a fixed size.
+#' - while vector of `value`, `min`, and `max` allows histogram bins dynamic adjustment.
```
vedhav, line 21: `length` as a variable is not declared so using it might cause confustion.
I liked the previous version when it was clear that the values passed should be `c(value, min, max)`
As a document reader, it makes it easier for me to write code after seeing that instead of reading the meaning and coming up with `c(...)`.

```suggestion
#' - When the length of `bins` is one: The histogram bins will have a fixed size based on the `bins` provided.
#' - When the length of `bins` is three: The histogram bins are dynamically adjusted based on `c(value, min, max)`.
```

vedhav (commented):
Just one comment. Else, LGTM.

R/tm_g_distribution.R
```diff
@@ -1,33 +1,33 @@
...
+#' It offers several tools, such as histograms, Q-Q plots, and various statistical tests to
+#' visually and statistically analyze the variable's distribution.
 #'
 #' @inheritParams teal::module
 #' @inheritParams teal.widgets::standard_layout
 #' @inheritParams shared_params
 #'
 #' @param dist_var (`data_extract_spec` or `list` of multiple `data_extract_spec`)
-#'  Variable to consider for the distribution analysis.
+#' Variable(s) for which the distribution  to be analyzed.
```
vedhav, line 12:
```suggestion
#' Variable(s) for which the distribution will be analyzed.
```

R/tm_g_distribution.R
kartikeyakirar: I've updated the c(..) usage in all the pull requests due to a review comment that I can't locate now. To maintain consistency, let's stick with this change.

I will incorporate rest of the change.

vedhav: Got it. Then I'll resolve and we can merge this one in.

PR #673, opened 2024-02-22
658 remove var labels@pre release@main

### LINKED ISSUE #658: replace `var_labels` with `teal.data::col_labels`
@insightsengineering/nest-core-dev 
`var_labels` (originally from `formatters`) was copied to several packages, including `teal.data`, where it was renamed to `col_labels` improved. In light of that, should `teal.modules.general` continue to use its copy of `var_labels` or rather switch to `teal.data::col_labels`?

It would require adding `teal.data` to `Imports`. ~~Currently `teal.data` is not a dependency (even though the package does rely on the `teal_data` class), we only use `eval_code` from `teal.code`.~~
_EDIT:_ `teal.data` was listed in `Suggests` but that was wrong and it is being moved to `Imports` in #643.

chlebowa:
Closes #658 

All instances of `var_labels` were replaced with `teal.data::col_labels`.
Source for `var_labels` was removed.


kartikeyakirar (approved):
LGTM 🚀

PR #675, opened 2024-02-23
`tm_g_regression` labels are no longer allowed out of bounds

### LINKED ISSUE #66: tm_a_regression now needs option to extend x axis range
With the implementation of outlier labels on the plot, the labels may go out of the default x range of the plot:

image

====>

image

Provenance: 
```
Creator: junlue
```

averissimo:
# Pull Request

Fixes #66

### Changes description

- Adds dependency to imports
- Uses `ggrepel` to generate the labels for outliers
- Adds option to `tm_g_regression` and encoding panel for size of segment

### Reviewer considerations

- Is the new encoding option necessary?
- This could be implemented as an optional dependency
  - with `if (requireNamespace("ggrepel", quietly) { <ggrepel> } else { <old behaviour> }`
  - In practice it's never an optional dependency as `ggmosaic` depends directly on it (`with Imports`)


averissimo:
image

averissimo:
Custom option on line segment might be overkill, I added it because it's a new visual element

image

chlebowa (changes requested):
>1. Is the new encoding option necessary?

Doesn't hurt. I suppose some control may be called for, depending on the data. Then again, the app dev must set the value to suit the current dataset. But can that be avoided altogether? Can `ggrepel` defaults be trusted? I would guess not.

> 2. This could be implemented as an optional dependency
with if (requireNamespace("ggrepel", quietly) { <ggrepel> } else { <old behaviour> }
In practice it's never an optional dependency as ggmosaic depends directly on it (with Imports)

Dependency status in `tmg` should reflect the package's role in `tmg` lone, irrespective of its ties to other packages. With the current implementation `ggrepel` shoudl be in `Imports`.
However, an optional dependency is reasonable as only one module uses it, which is the rule of thumb in `tmg`. 
How about making the new argument default to `NULL` and adding `if (!requireNamespace("ggrepel", quietly) stop("please install ggrepel")`? It's less code and in line with other cases.

R/tm_a_regression.R
```diff
@@ -22,6 +22,10 @@
 #' 5. Cook's distance
 #' 6. Residuals vs Leverage
 #' 7. Cook's dist vs Leverage
+#' @param label_segment_threshold (`numeric`) If scalar then the minimum distance
```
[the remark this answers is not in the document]
averissimo: I've mostly used the template of the `size` parameter, but I should've upgraded it.

I think I implemented 1-5, plus snuck in some missing assertions

chlebowa (changes requested):
:+1: 
Looks good to me. I would only ask that you reorder argument checks to match the order of formal arguments and unify checks of 3-number arguments, e.g. `alpha` and `plot_width`.

chlebowa (approved):
wrong button

averissimo:
The best way to unify the checks was to create a helper function.

It ensures the structure across arguments and it's the easiest to read.

R/utils.R
```diff
@@ -348,3 +348,30 @@ is_tab_active_js <- function(id, name) {
...
+    len = 3, any.missing = FALSE, finite = finite, null.ok = null.ok
+  )
+  is_bounded <- is_numeric && checkmate::test_number(
+    value[[1]],
+    lower = max(lower, value[[2]]),
+    upper = min(upper, value[[3]]),
+    null.ok = null.ok,
+    finite = finite
+  )
+  if (isFALSE(is_numeric) || isFALSE(is_bounded)) {
```
averissimo, line 368: Technically we could only test if `is_bounded` is FALSE, but I think this adds readability.

`is_numeric && checkmate::test_number(...)` is being used a few lines before so that we can safely do `value[[2]]` and `value[[3]]`

averissimo:
Asking for a re-review as it's a big change

ps. This could be applied to other functions during `pre-release` activities or after :-)

averissimo:
@chlebowa please ignore my previous comments since your approval.

I've reverted the changes and did what you asked, [these are the differences since](https://github.com/insightsengineering/teal.modules.general/pull/675/files/32b85429af2d171a8165ddeb7723daf4bbc972f4..d692cbc09d48e9f0e9cfe1761b8c46e693296c48)

(I'll create a new issue with my proposal of the `check_range_slider`)

R/tm_a_regression.R
```diff
@@ -23,13 +23,15 @@
...
+#' Minimum distance between label and point on the plot that triggers the creation of
+#' a line segment between the two.
+#' This may happen when the label cannot be placed next to the point as it overlaps another
+#' label or point.
 #' The value is used as the `min.segment.length` parameter to the [ggrepel::geom_text_repel()] function.
 #'
 #' It can take the following forms:
-#' - `numeric(1)`: It is a fixed valued used for the minimum distance and a slider is not presented in the UI.
-#' - `numeric(3)`: A slider is presented in the UI (under "Plot settings") to adjust the minimum distance dynamically.
+#' - `numeric(1)`: Fixed valued used for the minimum distance and the slider is not presented in the UI.
```
chlebowa, line 33:
```suggestion
#' - `numeric(1)`: Fixed value used for the minimum distance and the slider is not presented in the UI.
```

R/tm_a_regression.R
```diff
@@ -23,13 +23,15 @@
...
+#' - `numeric(3)`: The slider is presented in the UI (under "Plot settings") to adjust the minimum distance dynamically.
```
chlebowa, line 34:
```suggestion
#' - `numeric(3)`: A slider is presented in the UI (under "Plot settings") to adjust the minimum distance dynamically. `label_segment_threshold` is passed to the `value_min_max` argument in `teal.widgets::optionalSliderInputValMinMax`.
```

PR #691, opened 2024-02-27
Standardise function titles similar to what we have in tmc

### LINKED ISSUE #684: [Docs] Update the function doc titles to use the same format that we use in tmc
[tmc function references](https://insightsengineering.github.io/teal.modules.clinical/latest-tag/reference/index.html) follows this format for the function titles:

```r
teal Module: xxxxx xxxxx xxxx
```

It would be nice to follow the same format in tmg as well.

vedhav:
Closes #684 

Changes:

Titles now follow this format:
```
`teal` module: Function title
```

Additionally, moved the function params `alpha` and `size` to `shared_params` as they were used in 3 places.
Note that the `tm_g_bivariate()` uses `size` param with a different meaning. We might want to think about function arg standardization in the future.

### DIFF
--- a/R/tm_a_pca.R
+++ b/R/tm_a_pca.R
@@ -1,4 +1,4 @@
-#' Principal component analysis module
+#' teal Module: Principal Component Analysis
 #'
 #' Module conducts principal component analysis (PCA) on a given dataset and offers different
 #' ways of visualizing the outcomes, including elbow plot, circle plot, biplot, and eigenvector plot.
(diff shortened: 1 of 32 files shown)


R/tm_a_pca.R
```diff
@@ -1,4 +1,4 @@
-#' Principal component analysis module
+#' teal Module: Principal Component Analysis
```
averissimo, line 1: - Are we bringing back title case?
  - We've been avoiding them across all other `{teal.*}` packages. `{tmc}` is the outlier
- Backticks on `` `teal` ``?

```suggestion
#' `teal` module: Principal component analysis
```

or 

```suggestion
#' `teal` module: Principal Component Analysis
```

averissimo: I get it that we want to make it similar to `{tmc}`, but maybe we should only add the prefix and keep the same styling otherwise.

And suggest this to `{tmc}`

R/tm_g_association.R
```diff
@@ -1,4 +1,4 @@
-#' Stack plots of variables and show association with reference variable
+#' teal Module: Stack Plot with Association
```
averissimo, line 1: Do you think that association alone is enough to describe this module?

Consider reverting to old _(long)_ title

```suggestion
#' teal Module: Stack Plots of Variables and Show Association with Reference Variable
```

vedhav: Thanks! I changed it back now.

R/tm_a_pca.R
m7pr: @vedhav maybe you can also do a standarization in `{tmc}` while you are on a cleaning spree of titles in here?

vedhav: Sure thing

vedhav: Perhaps we can involve the sme team in the discussion before we decide on a format?

m7pr: yes. what about starting a discussion in this repository and CCing @shajoezhu @Melkiades and @edelarua

vedhav: Started the discussion [here](https://github.com/insightsengineering/teal.modules.clinical/discussions/1086)

Issue #674, opened 2024-02-23
Standardize the notation for `optional` parameters in documentation


kartikeyakirar:
Based on the discussions regarding the documentation style for indicating optional parameters within pacakge ([see details ](https://github.com/insightsengineering/teal.modules.general/pull/646#discussion_r1499323682)), it has been suggested that we adopt a unified approach to enhance consistency.

The proposed standard for denoting optional parameters is to follow the format:
``@param (`character`) optional ...``

averissimo:
An _"edge case"_ when the type is part of the description should be considered (@chlebowa first mentioned this that comment)

I've tentatively rephrased it like this: (\*\*optional\*\* surrounded by ** for emphasis)

WDYT?

```diff
-#' @param input_path optional, (`list`) of the input paths to either: specific files of accepted formats,
+#' @param input_path (`list`) of the input paths, **optional**. Each element can be:
```

Any suggestions on cases like this?

```diff
-#' @param row_facet optional, (`data_extract_spec` or `list` of multiple `data_extract_spec`)
-#'   Which data columns to use for faceting rows.
# a)
+#' @param row_facet (`data_extract_spec` or `list` of multiple `data_extract_spec`)
+#' **optional** specification of the data variable(s) to use for faceting rows.
# or b)
+#' @param row_facet (`data_extract_spec` or `list` of multiple `data_extract_spec`) **optional**.
+#' Specification of the data variable(s) to use for faceting rows.
```

kartikeyakirar:
IMO option a) might be more preferable because it integrates the "optional" keyword with the type description . and slightly maintain consistency with `input_path` example you suggested.

m7pr:
@insightsengineering/nest-core-dev would we like to tackle this before the release or after?

m7pr:
ah ok I just saw the PR, so probably before the release https://github.com/insightsengineering/teal.modules.general/pull/698

PR #698, opened 2024-02-28
standardize the notation for optional parameters

kartikeyakirar:
fixes https://github.com/insightsengineering/teal.modules.general/issues/674

### DIFF
--- a/R/tm_a_pca.R
+++ b/R/tm_a_pca.R
@@ -8,8 +8,8 @@
 #' @inheritParams teal::module
 #' @inheritParams shared_params
 #' @param dat (`data_extract_spec` or `list` of multiple `data_extract_spec`)
-#'   specifying columns used to compute PCA.
-#' @param font_size (`numeric`, optional) Specifies font size.
+#' specifying columns used to compute PCA.
+#' @param font_size (`numeric`) optional specifies font size.
 #' It controls the font size for plot titles, axis labels, and legends.
 #' - If vector of `length == 1` then the font sizes will have a fixed size.
 #' - while vector of `value`, `min`, and `max` allows dynamic adjustment.
--- a/R/tm_a_pca.R
+++ b/R/tm_a_pca.R
@@ -9,7 +9,7 @@
 #' @inheritParams shared_params
 #' @param dat (`data_extract_spec` or `list` of multiple `data_extract_spec`)
 #' specifying columns used to compute PCA.
-#' @param font_size (`numeric`) optional specifies font size.
+#' @param font_size (`numeric`) optional, specifies font size.
 #' It controls the font size for plot titles, axis labels, and legends.
 #' - If vector of `length == 1` then the font sizes will have a fixed size.
 #' - while vector of `value`, `min`, and `max` allows dynamic adjustment.
--- a/R/tm_a_regression.R
+++ b/R/tm_a_regression.R
@@ -14,8 +14,8 @@
 #' Regressor variables from an incoming dataset with filtering and selecting.
 #' @param response (`data_extract_spec` or `list` of multiple `data_extract_spec`)
 #' Response variables from an incoming dataset with filtering and selecting.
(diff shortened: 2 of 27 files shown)


R/tm_a_regression.R
```diff
@@ -14,8 +14,8 @@
 #' Regressor variables from an incoming dataset with filtering and selecting.
 #' @param response (`data_extract_spec` or `list` of multiple `data_extract_spec`)
 #' Response variables from an incoming dataset with filtering and selecting.
-#' @param default_outlier_label (`character`, optional) The default column selected to label outliers.
-#' @param default_plot_type (`numeric`, optional) Defaults to Response vs Regressor.
+#' @param default_outlier_label (`character`) optional default column selected to label outliers.
+#' @param default_plot_type (`numeric`) optional defaults to "Response vs Regressor".
```
m7pr, line 18: I'm just guessing - should that be in backticks?
```suggestion
#' @param default_plot_type (`numeric`) optional defaults to `"Response vs Regressor"`.
```

R/tm_a_pca.R
```diff
@@ -8,8 +8,8 @@
 #' @inheritParams teal::module
 #' @inheritParams shared_params
 #' @param dat (`data_extract_spec` or `list` of multiple `data_extract_spec`)
-#'   specifying columns used to compute PCA.
-#' @param font_size (`numeric`, optional) Specifies font size.
+#' specifying columns used to compute PCA.
+#' @param font_size (`numeric`) optional specifies font size.
```
vedhav, line 12:
```suggestion
#' @param font_size (`numeric`) optionally specifies font size.
```
Perhaps optionally would be better to use over here.

m7pr: I think that 

```r
(`numeric`) (optional) specifies font size.
```

would be the most natural in this situation : D

m7pr: or maybe 

```r
(`numeric`) optional, specifies font size.
```

vedhav: I like the second version from Marcin

kartikeyakirar: I believe we should standardize this and come to a decision. Do we incorporate it into the sentence directly, or should we use a specific notation? 


 Regarding the use of optionally or (option) create more formats so I think we should finalize it here first. https://github.com/insightsengineering/nestdevs-tasks/issues/52.

m7pr: just put a `,` after optional, so the sentence have logical sense. and it's all good.

chlebowa:
Just a quick note. The fact that an argument is optional is readily apparent in the function definition (no default value) so there is no need to use the word ever time. We don't always state the default value in the argument description, after all, only when it matters.

R/tm_a_regression.R
kartikeyakirar: its numeric type  and `Response vs Regressor` is label not choices that why I did not put backtick
image

m7pr:
If we are sure that the parameter does not get set inside functions body to anything (if it's NULL/empty during the call) then maybe it's a good idea to remove `optional` completely.

kartikeyakirar:
> The fact that an argument is optional is readily apparent in the function definition (no default value) 

I am confused,  Arguments are optional when you do not have to specify a value for them. They can have a default value, which is used if you do not specify.

chlebowa:
@kartikeyakirar You are correct, my brain flipped the logic.

R/tm_a_pca.R
kartikeyakirar: [c9ccb2c](https://github.com/insightsengineering/teal.modules.general/pull/698/commits/c9ccb2ce4224ac33291abbcedc7611152257fc30)

R/tm_a_regression.R
m7pr: ok

m7pr (approved):

Issue #701, opened 2024-02-29
[Bug]: `tm_g_bivariate` crash with generic example when either `x` or `y` are deselected


averissimo:
Error:

```r
Warning: Error in teal::validate_has_data: Assertion on 'x' failed: Must be of type 'data.frame', not 'double'.
  147: <Anonymous>
  146: stop
  145: mstop [/tmp/RtmpuH1W7o/R.INSTALL884135d00cc5/checkmate/R/helper.R#2]
  144: makeAssertion [/tmp/RtmpuH1W7o/R.INSTALL884135d00cc5/checkmate/R/makeAssertion.R#44]
  143: checkmate::assert_data_frame [/home/sandbox/staged_deps_refactor/renv/staging/1/checkmate#1]
  142: teal::validate_has_data [/tmp/Rtmp4sGw30/R.INSTALLa4637c62e002/teal/R/validations.R#47]
  141: <reactive:output_q> [/home/teal.modules.general/R/tm_g_bivariate.R#598]
  125: output_q [/tmp/RtmpkziAl8/R.INSTALL620331194582/shiny/R/reactives.R#870]
  124: <reactive> [/home//teal.modules.general/R/tm_g_bivariate.R#679]
  108: plot_r [/tmp/RtmpkziAl8/R.INSTALL620331194582/shiny/R/reactives.R#870]
   99: renderUI [/tmp/RtmpPvGI87/R.INSTALL3ae2652def2d/teal.widgets/R/plot_with_settings.R#412]
   98: func [/tmp/RtmpkziAl8/R.INSTALL620331194582/shiny/R/utils.R#1455]
   85: renderFunc [/tmp/RtmpkziAl8/R.INSTALL620331194582/shiny/R/shinywrappers.R#263]
   84: output$teal-main_ui-root-bivariate_plots-module-myplot-plot_out_main [/tmp/RtmpkziAl8/R.INSTALL620331194582/shiny/R/shinywrappers.R#133]
    3: runApp [/tmp/RtmpkziAl8/R.INSTALL620331194582/shiny/R/runapp.R#388]
    2: print.shiny.appobj [/tmp/RtmpkziAl8/R.INSTALL620331194582/shiny/R/shinyapp.R#553]
    1: <Anonymous>
Warning: Error in teal::validate_has_data: Assertion on 'x' failed: Must be of type 'data.frame', not 'double'.
[...]

PR #702, opened 2024-02-29
701 fixing the validation issue for x/y variable selection in `tm_g_bivariate`.

kartikeyakirar:
fixes https://github.com/insightsengineering/teal.modules.general/issues/701


chlebowa (changes requested):
There was a purpose to that function. Now you will have two messages where one would suffice.

kartikeyakirar:
Oh, I misread it and thought both selections were necessary. Simply adding `drop = FALSE `would be enough

PR #703, opened 2024-03-01
Removes unnecessary `# nolint ` after `object_name_linter` was changed

averissimo:
# Pull Request

- Part of #624
- Extends ANL rule to allow `ANL_OUTLIERS` and others variables that are prefixed with `ANL_` (in all caps)


m7pr (approved):
a legendary PR

PR #705, opened 2024-03-04
using teal.data datasets to cut package size

gogonzo:
This addresses CRAN comments about size of the package. See `R CMD build` to compare sizes. Datasets used in examples are taken from `teal.data`


kartikeyakirar:
@gogonzo The images in the vignettes directory are larger than the data. Shouldn't we reduce their size?

image

image

gogonzo:
> @gogonzo The images in the vignettes directory are larger than the data. Shouldn't we reduce their size or remove it ?
> 
> image
> image

I guess images are better compressed by `R CMD build` - after removing data package size cut from 4.8MB to <1MB. Can you confirm size of the tar.gz before my commit and after?

I can also remove images if @donyunardi is fine with it.

chlebowa:
> I guess images are better compressed by `R CMD build` - after removing data package size cut from 4.8MB to <1MB. Can you confirm size of the tar.gz before my commit and after?

How?? The images are < 1 MB.

gogonzo:
Closing this as @donyunardi has already resubmitted a package with cut size.

chlebowa:
Removing the datasets from `tmg` and using ones from `teal.data` makes sense anyway. I suggest merging this.

gogonzo:
> Removing the datasets from `tmg` and using ones from `teal.data` makes sense anyway. I suggest merging this.

@donyunardi has already submitted the package. If they reject for some reason I can resurrect this PR

Issue #712, opened 2024-03-22
Implement shinytest2 for tmg


donyunardi:
This is a continuation of #503

Using the shinytest2 helper class that we built for `teal`, let extend the feature for tmg. 

We have 15 teal module functions in tmg

### Test specs
- [ ] Test if the example apps for the modules tm_* can be run without shiny errors

averissimo:
Created an initial PR that will support {tmg} testing (#714).

I'm wondering about what datasets should be used as the base for automated testing, (1) iris/mtcars/<general> or (2) CDISC

CDISC is closer to the overall goal of teal, on the other hand, all the examples modules should work with general data, just as it was recently added to all the examples (a mix between iris/mtcars/CO2/USArrests/...)

Any thoughts folks?

kartikeyakirar:
I believe general datasets can encompass all necessary functionalities since (tmg) is designed as standard modules for data operations. 

I prefer using general datasets to maintain simplicity, except in cases where specific functionalities are only relevant for CDISC datasets

averissimo:
>  - [ ] Test if the example apps for the modules tm_* can be run without shiny errors

I think we can focus our effort partially on merging https://github.com/insightsengineering/teal.modules.clinical/pull/983 and then copy/paste it to {tmg}

This will allow our automated tests to be focused on the individual module's functionality (and leave basic error detection to this effort that is 2-in-1

averissimo:
The automated tests being developed for {tmc} works in this package as well with 2 exceptions `[ FAIL 2 | WARN 0 | SKIP 0 | PASS 68 ]`

Require custom input at startup (selections are not available on the module's parameters as of now)
- `tm_file_viewer` (file being chosen)
- `tm_g_distribution`: choosing a test on a collapsed section of encoding

It lives on a tentative branch as of now ([`477_shinytest2_examples@main`](https://github.com/insightsengineering/teal.modules.general/tree/477_shinytest2_examples%40main))

We could address this in different ways:
- Exclude example app and create custom test
- Set missing options via module call (adding new argument)
- Allow for some shiny validation errors based on a regular expression

pawelru:
Thanks for running this in here. Out of the options you listed I think I like allow-list (for warnings / validation / etc.) the most and disliked adding new argument the most :) It would be great to have such list very specific and accurate, e.g. for file XYZ accept validation with ID ABC and warning DEF.

averissimo:
[...]

PR #716, opened 2024-04-10
712 - `{shinytest2}` for `tm_a_pca`

averissimo:
# Pull Request

Part of #712

### Changes description

- Adds 2 new e2e tests using `shinytest2`
  - Data extract inputs
    - [x] Main
    - [x] Available on specific plots
  - Encoding options via "(guided) monkey typing"
- Fixes typo on module

### Considerations

- End-2-End tests are complex and require complex set of expectations
  - Otherwise, we risk having a very long testing pipeline (`AppDriver` has a relevant start-up time)
- How complete do we want to be on the encoding testing?
  - Took a brute force approach here, but small changes/removals will break the tests

PR #717, opened 2024-04-11
712 - `{shinytest2}` for `tm_front_page`

kartikeyakirar:
Part of #712

- [x] tested header-text
- [x] tested table displayed correctly
- [x] additional argument tags.
- [x] checked metadata data


tests/testthat/test-shinytest2-tm_front_page.R
```diff
@@ -0,0 +1,65 @@
...
+        tables = list("MTCARS" = head(mtcars, 5), "IRIS" = head(iris, 5)),
+        additional_tags = HTML("Additional HTML or shiny tags go here <br>"),
+        footnotes = c("X" = "is the first footnote", "Y is the second footnote"),
+        show_metadata = TRUE
+      )
+    ),
+    timeout = 3000
+  )
+
+  app$wait_for_idle()
```
averissimo, line 27: This is not necessary due to https://github.com/insightsengineering/teal/blob/05c0775fa56f85496f582fc8dc16bc506e88c89c/R/TealAppDriver.R#L69

```suggestion
```

tests/testthat/test-shinytest2-tm_front_page.R
```diff
@@ -0,0 +1,65 @@
...
+  app$expect_no_shiny_error()
+
+  # header text
+  testthat::expect_identical(
+    rlang::hash(app$get_html("#front_page_headers")),
+    "2b2ab743c68213927d0578b7e39f5873"
+  )
```
averissimo, line 34: Instead of a hash, how about using `app$expect_html(#front_page_headers)`?

It would be easier to update

The same applies to the other functions with a mix of existing `app$expect_*` and `testthat::expect_snapshot()` for the `app$get_active_module_output`

Unless you want to create this method first upstream at `{teal}`

averissimo (changes requested):
Using snapshot tests might be better than `rlang::hash` for maintanability

tests/testthat/test-shinytest2-tm_front_page.R
kartikeyakirar: done [15e409c](https://github.com/insightsengineering/teal.modules.general/pull/717/commits/15e409cbd7648dcef1903d5b859f552a14c6dbc7)

tests/testthat/test-shinytest2-tm_front_page.R
```diff
@@ -0,0 +1,48 @@
+test_that("e2e: tm_front_page initializes without errors", {
+  skip_if_too_deep(9)
```
vedhav, line 2:
```suggestion
  skip_if_too_deep(5)
```
The depth can only be a number between 1-5. Please change it to 5.

kartikeyakirar: my bad, thanks

tests/testthat/test-shinytest2-tm_front_page.R
```diff
@@ -0,0 +1,48 @@
...
+
+  app$expect_screenshot(selector = app$active_module_element("table_2"))
+  testthat::expect_match(app$get_active_module_output("table_2"), "IRIS")
+
+
+  # additional tags
+  app$expect_html("#front_page_custom_html")
+
+  # show metadata
+  app$click(NS(app$active_module_ns(), "metadata_button"))
```
vedhav, line 44: I get a weird error when this click is performed.
```r
Error in `app_wait_for_idle()`:
! An error occurred while waiting for Shiny to be stable
Run `rlang::last_trace()` to see where the error occurred.
```

tests/testthat/test-shinytest2-tm_front_page.R
```diff
@@ -0,0 +1,48 @@
...
+    timeout = 3000
+  )
+
+  app$expect_no_shiny_error()
+
+  # header text
+  app$expect_html(selector = "#front_page_headers")
+
+  # tables
+  app$expect_screenshot(selector = app$active_module_element("table_1"))
```
vedhav, line 33: I do not like when we perform the snapshot tests. They are very fragile. Let's try to find some other way to test this.

vedhav: I see that we moved to snapshot tests after using the hash. They are better to debug compared to hash. Both are fragile. I've already seen TLG catalog snapshot updates after the calculation change being made.
I think we can use them in cases where we're certain that the output will not change, what we have here seems good to me.
Do we need to add the snapshots for all three platforms in the PR? Now it only contains windows and R 4.3
Please also consider using a [seed](https://rstudio.github.io/shinytest2/reference/AppDriver.html#method-new-) during the `TealAppDriver` init when we use snapshots.

tests/testthat/test-shinytest2-tm_front_page.R
kartikeyakirar: I did not get this error. should we add app_wait_for_idle() after click ?

tests/testthat/test-shinytest2-tm_front_page.R
kartikeyakirar: do you prefer if we just do `expect_html `as they are HTML elements.

tests/testthat/test-shinytest2-tm_front_page.R
```diff
@@ -0,0 +1,48 @@
+test_that("e2e: tm_front_page initializes without errors", {
+  skip_if_too_deep(5)
+  require(shinytest2)
+
+  data <- simple_cdisc_data()
```
m7pr, line 5: can you prefix `simple_cdisc_data` ? does it come from tmg or any other package?

tests/testthat/test-shinytest2-tm_front_page.R
```diff
@@ -0,0 +1,48 @@
...
+  data <- within(data, {
+    attr(ADSL, "metadata") <- list("Author" = "NEST team", "data_source" = "synthetic data")
+  })
+
+  app <- TealAppDriver$new(
```
m7pr, line 10:
```suggestion
  app <- teal::TealAppDriver$new(
```

tests/testthat/test-shinytest2-tm_front_page.R
```diff
@@ -0,0 +1,48 @@
...
+    data = data,
+    modules = teal::modules(
```
m7pr, line 12: does it need to be covered in `modules`? 
```suggestion
    modules = 
```

tests/testthat/test-shinytest2-tm_front_page.R
vedhav: It's a test helper

tests/testthat/test-shinytest2-tm_front_page.R
m7pr: maybe `$expect_html` and verify `$get_text()`

tests/testthat/test-shinytest2-tm_front_page.R
kartikeyakirar: this is helper function defined in  helper-TealAppDriver.R

tests/testthat/test-shinytest2-tm_front_page.R
kartikeyakirar: `TealAppDriver` is internal class .  we have import non-exported `TealAppDriver `from `teal` package in  helper-TealAppDriver.R

tests/testthat/test-shinytest2-tm_front_page.R
m7pr: Screenshots not only area prone to R version and platform, but also to system fonts : ) 

image

m7pr: @vedhav, is snapshot and screenshot something else? Kartik had screenshots (images) in here, and snaphosts are stored JSONs?

tests/testthat/test-shinytest2-tm_front_page.R
m7pr: `TealAppDrvier$click()` already have `app$wait_for_idle()`

tests/testthat/test-shinytest2-tm_front_page.R
kartikeyakirar: Not for a single module. I will remove it .

tests/testthat/test-shinytest2-tm_front_page.R
m7pr: oh my, this hash idea seems an overkill : P

tests/testthat/test-shinytest2-tm_front_page.R
m7pr: ah, just seeing this in `tests/testthat/helper-TealAppDriver.R`

tests/testthat/test-shinytest2-tm_front_page.R
m7pr: also remove the closing bracket

tests/testthat/test-shinytest2-tm_front_page.R
vedhav: I was referring snapshot to as any kind of "store the output -> check it later" which includes images, json or any other file format.
They are great for regression testing, where we need to make sure that the UI does not change, shinytest2 is primarily meant for these tests. At least we are testing just a particular selector area for screenshots so we're okay. But, I still don't like them.

averissimo: There is no random-ness on this module so I think it plays well here.

It tracks the specific tables instead of the whole app or figures.

Do you have an alternative in mind?

tests/testthat/test-shinytest2-tm_front_page.R
kartikeyakirar: I mean if problem persists we can increase time.

tests/testthat/test-shinytest2-tm_front_page.R
kartikeyakirar: done https://github.com/insightsengineering/teal.modules.general/pull/717/commits/8c6be6159dd2a64eb3d2dfbfb90fa4d5078ce6c0

tests/testthat/test-shinytest2-tm_front_page.R
averissimo: I can update on ubuntu

tests/testthat/test-shinytest2-tm_front_page.R
averissimo: I had the same problem and it was a missing dependency that was causing an issue `rvest`.

Can you see if that solves your problem @vedhav ? if so, we will need to add it to the `Suggests` of `{tmg}`

averissimo: For context, `{tmg}` has no reason to install it as it's on Suggests of a dependency.. that is: `{teal}`

tests/testthat/test-shinytest2-tm_front_page.R
```diff
@@ -0,0 +1,40 @@
...
+
+test_that("e2e: tm_front_page displays metadata", {
+  skip_if_too_deep(5)
+  app_driver <- app_driver_tm_front_page()
+  # show metadata
+  app_driver$click(NS(app_driver$active_module_ns(), "metadata_button"))
+  app_driver$is_visible(selector = app_driver$active_module_element("metadata_table"))
+
+  app_driver$stop()
+})
```
m7pr, line 40: @kartikeyakirar is metadata somethnig specific to the module or something specific to `{teal}`? if it's specific to teal page please move this test to `{teal}`.

kartikeyakirar: There is `show_metadata = TRUE` parameter in module  to show metadata and its not specific to teal.

averissimo (approved):
LGTM!

Consider moving the `app_driver_tm_front_page` function definition to the test file as suggested [here](https://github.com/insightsengineering/teal.modules.general/pull/716#discussion_r1562407079)

kartikeyakirar:
@averissimo done here [8340a5c](https://github.com/insightsengineering/teal.modules.general/pull/717/commits/8340a5c5bcf81f1f834c404e0c168b924f8eb565)

averissimo (approved):

PR #730, opened 2024-04-16
712 - `{shinytest2}` for `tm_outliers`

m7pr:
Part of #712


m7pr:
Hey @kartikeyakirar while reviewing, would you be able to check why this test fails

```r
── Failure: e2e - tm_outliers:
  data extract spec elements are initialized with the default values
  specified by outlier_var and categorical_var argument ──
app_driver$get_active_module_input("categorical_var-dataset_CO2_singleextract-filter1-vals") not identical to c(...).
target is NULL, current is character

Error:
! Test failed
Backtrace:
    ▆
 1. ├─testthat::test_that(...)
 2. │ └─withr (local) `<fn>`()
 3. └─reporter$stop_if_needed()
 4.   └─rlang::abort("Test failed", call = NULL)
```

tests/testthat/test-shinytest2-tm_outliers.R
```diff
@@ -0,0 +1,312 @@
...
+  app_driver$expect_no_shiny_error()
+
+  testthat::expect_identical(
+    app_driver$get_active_module_input("categorical_var-dataset_CO2_singleextract-filter1-col"),
+    "Plant"
+  )
+  app_driver$set_active_module_input("categorical_var-dataset_CO2_singleextract-filter1-col", "Type")
+  app_driver$expect_no_shiny_error()
+
+  testthat::expect_identical(
```
kartikeyakirar, line 92: for different order we could use expect_setequal?

tests/testthat/test-shinytest2-tm_outliers.R
```diff
@@ -0,0 +1,312 @@
...
+    app_driver$get_active_module_input("categorical_var-dataset_CO2_singleextract-filter1-vals"),
```
kartikeyakirar, line 93: this test is failing because of setting `Type` in filter by.
`app_driver$set_active_module_input("categorical_var-dataset_CO2_singleextract-filter1-col", "Type")`

I would suggest we should test default values separately or first and then set values.

kartikeyakirar: image

tests/testthat/test-shinytest2-tm_outliers.R
```diff
@@ -0,0 +1,312 @@
...
+
+  app_driver <- app_driver_tm_outlier()
+  app_driver$expect_no_shiny_error()
+
+  testthat::expect_identical(
+    app_driver$get_active_module_input("outlier_var-dataset_CO2_singleextract-select"),
+    "uptake"
+  )
+  app_driver$set_active_module_input("outlier_var-dataset_CO2_singleextract-select", "conc")
+  app_driver$expect_no_shiny_error()
```
kartikeyakirar, line 83: I believe that the `app_driver$expect_no_validation_error()` test is better one  when the inputs have been set.

tests/testthat/test-shinytest2-tm_outliers.R
m7pr: I think it's fine right now, there is an issue to solve the order

tests/testthat/test-shinytest2-tm_outliers.R
m7pr: fixed - thanks ce4da519688047069959912a61625c36e2ee27a2

m7pr:
Thanks @kartikeyakirar I incorporated your feedback. Let me know if this is good to go

kartikeyakirar (approved):
Great Job 🚀

Issue #736, opened 2024-04-19
[Feature Request]: Allow app dev to supply card function.


chlebowa:
### Feature description

Card functions being hard coded preclude the app dev and app user from shaping the report.

If there were an argument for a card function, one could customize their report without having to duplicate and modify a module.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

donyunardi:
Thanks @chlebowa  
Definitely worth looking into this and it's great we have a head start with the PR that you raised.

Perhaps this feature can also encapsulate the request to [skip reporter for a single module](https://github.com/insightsengineering/teal/issues/847) and [disable them altogether](https://github.com/insightsengineering/teal/issues/871).

We are interested in this and looking forward to collaborate with you on this.
I've added this to the board for our next increment goal so we have time to review and discuss the design together.

chlebowa:
> Perhaps this feature can also encapsulate the request to [skip reporter for a single module](https://github.com/insightsengineering/teal/issues/847) and [disable them altogether](https://github.com/insightsengineering/teal/issues/871).

Reporting is built into modules and cannot be controlled by the module constructor (e.g. `tm_g_distribution`).
I can see setting the card function to `NA` as a signal to waive the built-in reporting but that would require other changes to the module. I don't think the issues overlap but perhaps this one can be a basis for the others.

I mentioned [here](https://github.com/insightsengineering/teal.modules.general/pull/737#discussion_r1574935910) a way of adding reporting to a module by a function call. If reporting were added explicitly, lie with `with_reporter(tm_g_distribution(...), card_function)` or `with_reporter(tm_g_distribution, card_function)(...)`, dropping reporting would be trivial.


> I've added this to the board for our next increment goal

Would that be June through July?

donyunardi:
> I can see setting the card function to NA as a signal to waive the built-in reporting but that would require other changes to the module. I don't think the issues overlap but perhaps this one can be a basis for the others.

That's exactly what I thought. I see that this can be the basis for other requirements. 

> Would that be June through July?

Correct, hope that's okay.

chlebowa:
We'll make it work 👌

m7pr:
[...]

PR #737, opened 2024-04-19
736 Allow custom card functions in modules

chlebowa:
Solves #736 

Allows app developers to pass card functions to modules as arguments.

### Demonstration

Example app demonstrating use of locally defined card function:
<details> <summary> example app </summary>

```
rm(list = ls())

library(teal)
devtools::load_all("./teal.modules.general")
library(teal.widgets)

local_card_function <- function(comment, label) {
  card <- teal::report_card_template(
    title = "Distribution Plot",
    label = label,
    with_filter = FALSE,
    filter_panel_api = filter_panel_api
  )
  card$append_text("Dummy Card", "header2")
  card$append_text(paste(
    "This report card was generated by a card function",
    "defined in the global environment."
  ))
  if (!comment == "") {
    card$append_text("Comment", "header3")
    card$append_text(comment)
  }
  card
}

data <- teal_data() |> within({
  iris <- iris
  mtcars <- mtcars
  faithful <- faithful
})
datanames(data) <- c("iris", "mtcars", "faithful")
filter <- teal_slices(
  teal_slice("iris", "Petal.Length", selected = c(1, 5.1)),
  teal_slice("iris", "Species", selected = c("setosa", "virginica")),
  teal_slice("mtcars", "cyl", selected = 6L),
  teal_slice("faithful", "waiting", selected = c(76, 82)),
  module_specific = TRUE,
  mapping = list(
    # "Distribution Module" = "iris Petal.Length",
    "Example" = "faithful waiting",
    "global_filters" = "iris Species"
  )
)
app <- init(
  data = data,
  modules = modules(
    tm_g_distribution(
      label = "Default card function",
      dist_var = data_extract_spec(
        dataname = "iris",
        select = select_spec(variable_choices("iris"), "Petal.Length")
      ),
      ggplot2_args = ggplot2_args(
        labs = list(subtitle = "Plot generated by Distribution Module")
      )
    ),
    tm_g_distribution(
      label = "Local card function",
      dist_var = data_extract_spec(
        dataname = "iris",
        select = select_spec(variable_choices("iris"), "Petal.Length")
      ),
      ggplot2_args = ggplot2_args(
        labs = list(subtitle = "Plot generated by Distribution Module")
      ),
      card_function = local_card_function
    ),
    example_module("Example")
  ),
  filter = filter
)

runApp(app, launch.browser = TRUE)
```

</details>

Also works with card function defined in another package (not shown).

### Rationale

Currently each module has a hard-coded card function so the module developer has full control over what is added to the report. Neither the app user, nor the app developer has any influence on this.

Being able to redefine the card function is a valid use case.

### Scoping Consideration

Current card functions are defined within the module server functions because they rely heavily on bindings that exist in the execution environment of the server function. In order for a function defined elsewhere to be able to access those bindings, a function is introduced, `hydrate_function`. `hydrate_function` changes the enclosing environment of a function (here, a card function) to bring another frame into scope. It can also add other bindings to the funciton's enclosure.

### Implementation

- Card functions from all modules that support reporting are isolated as separate, unexported functions. The naming convention is `<module_name>_card_function`.
- All modules that support reporting receive a `card_function` formal argument. If not provided, falls back to the corresponding card function for the module.
- Within modules, the card function, whether one provided in the call or the fallback value, is passed through `hydrate_function` to extend its scope to include the module server function's execution environment. This call is identical in all modules:
`card_function <- hydrate_function(card_function, with_filter, filter_panel_api)`.

### NOTE

This is a proof of concept, changes have only been made to `tm_g_distribution`.

### Progress

- [ ] tm_a_pca.R
- [ ] tm_a_regression.R       
- [ ] tm_data_table.R
- [ ] tm_g_association.R      
- [ ] tm_g_bivariate.R        
- [x] tm_g_distribution.R     
- [ ] tm_g_response.R
- [ ] tm_g_scatterplot.R      
- [ ] tm_g_scatterplotmatrix.R
- [ ] tm_missing_data.R       
- [ ] tm_outliers.R
- [ ] tm_t_crosstable.R       
- [ ] tm_variable_browser.R

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: teal.modules.general
 Title: General Modules for 'teal' Applications
-Version: 0.3.0.9034
-Date: 2024-05-16
+Version: 0.3.0.9038
+Date: 2024-06-27
 Authors@R: c(
     person("Dawid", "Kaledkowski", , "dawid.kaledkowski@roche.com", role = c("aut", "cre")),
     person("Pawel", "Rucki", , "pawel.rucki@roche.com", role = "aut"),
@@ -46,10 +46,10 @@ Imports:
     stringr (>= 1.4.1),
     teal.code (>= 0.5.0),
     teal.data (>= 0.5.0),
-    teal.logger (>= 0.1.3.9013),
+    teal.logger (>= 0.2.0.9004),
     teal.reporter (>= 0.3.0),
     teal.widgets (>= 0.4.0),
-    tern (>= 0.9.3),
+    tern (>= 0.9.5),
     tibble (>= 2.0.0),
     tidyr (>= 0.8.3),
     tools,
@@ -67,31 +67,34 @@ Suggests:
     jsonlite,
     knitr (>= 1.42),
     lattice (>= 0.18-4),
+    logger (>= 0.2.0),
     MASS,
     nestcolor (>= 0.1.0),
     pkgload,
-    rtables (>= 0.6.6),
+    rlang (>= 1.0.0),
+    rmarkdown (>= 2.23),
+    rtables (>= 0.6.8),
     rvest,
     shinytest2,
     sparkline,
-    testthat (>= 3.0.4),
+    testthat (>= 3.1.9),
     withr (>= 2.0.0)
 VignetteBuilder:
     knitr
 Config/Needs/verdepcheck: haleyjeppson/ggmosaic, tidyverse/ggplot2,
     rstudio/shiny, insightsengineering/teal,
     insightsengineering/teal.transform, mllg/checkmate, tidyverse/dplyr,
-    rstudio/DT, tidyverse/forcats, daroczig/logger, r-lib/scales,
-    daattali/shinyjs, shinyTree/shinyTree, rstudio/shinyvalidate,
-    dreamRs/shinyWidgets, tidyverse/stringr,
-    insightsengineering/teal.code, insightsengineering/teal.data,
-    insightsengineering/teal.logger, insightsengineering/teal.reporter,
-    insightsengineering/teal.widgets, insightsengineering/tern,
-    tidyverse/tibble, tidyverse/tidyr, tidymodels/broom,
-    daattali/colourpicker, daattali/ggExtra, aphalo/ggpmisc, aphalo/ggpp,
-    slowkow/ggrepel, baddstats/goftest, gridExtra, ramnathv/htmlwidgets,
-    jeroen/jsonlite, yihui/knitr, deepayan/lattice, MASS,
-    insightsengineering/nestcolor, r-lib/rlang,
+    rstudio/DT, tidyverse/forcats, r-lib/rlang, r-lib/scales, daattali/shinyjs,
+    shinyTree/shinyTree, rstudio/shinyvalidate, dreamRs/shinyWidgets,
+    tidyverse/stringr, insightsengineering/teal.code,
+    insightsengineering/teal.data, insightsengineering/teal.logger,
+    insightsengineering/teal.reporter, insightsengineering/teal.widgets,
+    insightsengineering/tern, tidyverse/tibble, tidyverse/tidyr,
+    tidymodels/broom, daattali/colourpicker, daattali/ggExtra,
+    aphalo/ggpmisc, aphalo/ggpp, slowkow/ggrepel, baddstats/goftest,
+    gridExtra, ramnathv/htmlwidgets, jeroen/jsonlite, yihui/knitr,
+    daroczig/logger, deepayan/lattice, MASS,
+    insightsengineering/nestcolor, r-lib/rlang, rstudio/rmarkdown,
     insightsengineering/rtables, tidyverse/rvest, sparkline,
     rstudio/shinytest2, insightsengineering/teal.data, r-lib/testthat,
     r-lib/withr
--- a/R/tm_g_distribution.R
+++ b/R/tm_g_distribution.R
@@ -364,7 +364,7 @@ srv_distribution <- function(id,
   checkmate::assert_class(data, "reactive")
   checkmate::assert_class(isolate(data()), "teal_data")
   moduleServer(id, function(input, output, session) {
-    if (shiny::isRunning()) logger::log_shiny_input_changes(input, namespace = "teal.modules.general")
+    teal.logger::log_shiny_input_changes(input, namespace = "teal.modules.general")
 
     setBookmarkExclude("params_reset")
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,9 +1,9 @@
-# teal.modules.general 0.3.0.9034
+# teal.modules.general 0.3.0.9038
 
 * Removed `Show Warnings` modals from modules.
 
 ### Enhancements
-* Added `logger` functionality for logging changes in shiny inputs in all modules. `logger` was added to Imports.
+* Added `teal.logger` functionality for logging changes in shiny inputs in all modules.
 
 * Users can now provide their own card functions to specify the content that modules send to reports.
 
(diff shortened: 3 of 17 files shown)


pawelru:
> Being able to redefine the card function is a valid use case.

I very much agree with this. I like the request and I do like the idea for implementation - that is: a card factory function. There are some details that I am still not quite sure yet and these could potentially require changes in `teal.reporter` itself (hopefully not but I cannot guarantee).

R/tm_g_distribution.R
```diff
@@ -1282,37 +1290,40 @@ srv_distribution <- function(id,
...
-      teal.reporter::simple_reporter_srv("simple_reporter", reporter = reporter, card_fun = card_fun)
+      card_function <- hydrate_function(card_function, with_filter, filter_panel_api)
+      teal.reporter::simple_reporter_srv("simple_reporter", reporter = reporter, card_fun = card_function)
     }
     ###
   })
 }
+
+#' @keywords internal
+tm_g_distribution_card_function <- function(comment, label) { #nolint: object_length.
```
pawelru, line 1301: my main concern is that this is not a true list of required arguments. We are using `with_filter` and `filter_panel_api`. Moreover, we are using `dist_r()`, `pws1` and many more object from the parent environment.
You have addressed this with `hydrate_function` but I think it would be better not to be in need of such functionality. I can think of a few ways how to do this:
- extend function arguments
- pass data as environment
- this being a module?

(I really hope that there is no strict check for these and only these argument somewhere in the teal.reporter).

Each has its pros and cons and we probably need to think more which one would be best for this task. Glad you stopped early to allow for a discussion like this.

Looking at the changes - this is how it was written in the past so I definitely not blaming you for this. This PR is a great opportunity to make it right.

chlebowa: > (I really hope that there is no strict check for these and only these argument somewhere in the teal.reporter).

[There is one, actually](https://github.com/insightsengineering/teal.reporter/blob/ad7fd2041f0d7cac83d0d97dbbe3ed7914a197ea/R/AddCardModule.R#L89). I considered just adding `...` to the card function's formals but that is forbidden. I think it would work but I am not sure about resolving conflicts between the variables passed to the ellipsis and bindings in the card function's parent environment. If we can figure out a way for the card functions to be pure, this would not be a problem.

I tried to limit the proposed changes to the module packages because I assumed modifying `teal.reporter` is off the table.

Note that passing the caller environment is not sufficient, at least in this module (I assume in others as well). `with_filter` and `filter_panel_api` are also required but they exist in the caller's parent frame, not in the caller itself.

I will be happy to discuss a satisfactory solution.

----

Note also that with the proposed solution the `if (with_reporter)` chunk would be the same in every module, which means _all code added to the module to enable reporting _would be the same in every module. The only difference would be the default card function, which is separate.

This in turn opens another possibility: have a function called `with_reporter` that would modify any module object by adding the code required for reporting into the module functions. Then adding support for reporting would be limited to something like `modules(tm_g_distribution(...) |> with_reporter(card_fun = foo))`. But that's a separate conversation.

pawelru: Ahh that's bad :( Then I would say that please feel free to modify this in `teal.reporter`. I feel that this is the true root-cause of the issue. I also think that this (relaxing this check on formals) would make creating card-factory-fun much simpler and also encapsulated so that it can be run without any additional steps - just provide what's required.
That's exactly what I'm looking for - more simplicity. I'm a little bit afraid of introducing a requirement of _hydrating_ a card-factory-fun. I don't think that most of our users (in particular: not developers) would handle this.

chlebowa: Have a look at this alternative: https://github.com/insightsengineering/teal.modules.general/pull/742

I don't think it is full proof because of the way `teal.reporter` checks and handles formal arguments of `card_fun`:
https://github.com/insightsengineering/teal.reporter/blob/ad7fd2041f0d7cac83d0d97dbbe3ed7914a197ea/R/AddCardModule.R#L166

Note that environment shenanigans are already built into `teal.reporter`:
https://github.com/insightsengineering/teal.reporter/blob/ad7fd2041f0d7cac83d0d97dbbe3ed7914a197ea/R/AddCardModule.R#L173

I think this branch is the superior solution (not necessary the ideal one).

chlebowa:
I have read the CLA Document and I hereby sign the CLA

DESCRIPTION
```diff
@@ -35,6 +35,8 @@ Imports:
     DT (>= 0.13),
     forcats (>= 1.0.0),
     grid,
+    logger (>= 0.3.0),
```
m7pr, line 38:
```suggestion
```

DESCRIPTION
```diff
@@ -82,7 +84,7 @@ VignetteBuilder:
 Config/Needs/verdepcheck: haleyjeppson/ggmosaic, tidyverse/ggplot2,
     rstudio/shiny, insightsengineering/teal,
     insightsengineering/teal.transform, mllg/checkmate, tidyverse/dplyr,
-    rstudio/DT, tidyverse/forcats, r-lib/scales, daattali/shinyjs,
+    rstudio/DT, tidyverse/forcats, r-lib/rlang, r-lib/scales, daattali/shinyjs,
```
m7pr, line 87: As `rlang` was introduced as a dependency, it also needs to be added here and in `.pre-commit-config.yaml`

R/utils.R
```diff
@@ -278,3 +282,56 @@ assert_single_selection <- function(x,
   }
   invisible(TRUE)
 }
+
+#' Hydrate a function's enclosing environment
+#'
+#' Add bindings of an environment to a function's parent environment.
+#'
+#' This allows any funciton to use bindings present in any environment
```
m7pr, line 290:
```suggestion
#' This allows any function to use bindings present in any environment
```

R/utils.R
```diff
@@ -278,3 +282,56 @@ assert_single_selection <- function(x,
...
+#' as if the funciton were defined there.
```
m7pr, line 291:
```suggestion
#' as if the function were defined there.
```

R/utils.R
```diff
@@ -278,3 +282,56 @@ assert_single_selection <- function(x,
...
+#' All bindings of the additional environment are added to the function's enclosure,
+#' except bindings existing in the enclosure are _not_ overwritten.
+#'
+#' One may also want to add variables that are not bound in the caller
+#' but are accessible from the caller, e.g. they exist in the caller's parent frame.
+#' This may happen in `shiny` modules because `moduleServer` is called
+#' by the module server function so the server funciton's arguments are in scope
```
m7pr, line 298:
```suggestion
#' by the module server function so the server function's arguments are in scope
```

R/tm_g_distribution.R
```diff
@@ -118,7 +118,8 @@ tm_g_distribution <- function(label = "Distribution Module",
                               plot_height = c(600, 200, 2000),
                               plot_width = NULL,
                               pre_output = NULL,
-                              post_output = NULL) {
+                              post_output = NULL,
+                              card_function) {
```
m7pr, line 122:
```suggestion
                              card_function = tm_g_distribution_card_function) {
```

I would expose this so that it's visible which function is used. However that would require exposing N reporting cards for N modules

R/tm_g_distribution.R
```diff
@@ -169,6 +170,12 @@ tm_g_distribution <- function(label = "Distribution Module",
 
   checkmate::assert_multi_class(pre_output, c("shiny.tag", "shiny.tag.list", "html"), null.ok = TRUE)
   checkmate::assert_multi_class(post_output, c("shiny.tag", "shiny.tag.list", "html"), null.ok = TRUE)
+
+  if (missing(card_function)) {
+    card_function <- tm_g_distribution_card_function
+  } else {
+    checkmate::assert_function(card_function)
+  }
```
m7pr, line 178: Then this can only be limited to an assert_function if we go with this approach https://github.com/insightsengineering/teal.modules.general/pull/737/files#r1671978213

R/tm_g_distribution.R
m7pr: Haven't yet checked the alternative, but in this case I would opt for relaxing teal.reporter checks to allow passing ellipsis which would simply the process and would not require the usage of hydrating in this case.

m7pr: passing `env` instead of `...` also works!

PR #733, opened 2024-04-17
712 - `{shinytest2}` for `tm_t_crosstable`

kartikeyakirar:
Part of https://github.com/insightsengineering/teal.modules.general/issues/712

PR #727, opened 2024-04-12
712 - `{shinytest2}` for `tm_missing_data`

kartikeyakirar:
Part of https://github.com/insightsengineering/teal.modules.general/issues/712

PR #738, opened 2024-04-22
fixing shinytest2 fail for tm_missing_data

kartikeyakirar:
A test in tm_missing_data is failing because a slider's default value gives an error. It passed in PR [727](https://github.com/insightsengineering/teal.modules.general/pull/727) with a value of 2L, but here it was showing 1L. I fixed it in PR [733](https://github.com/insightsengineering/teal.modules.general/pull/733) with app_driver$get_active_module_input("iris-combination_cutoff"), but it's causing an error in another PR again.

To fix I have set.seed in` teal_data()`

### DIFF
--- a/tests/testthat/test-shinytest2-tm_misssing_data.R
+++ b/tests/testthat/test-shinytest2-tm_misssing_data.R
@@ -1,5 +1,6 @@
 app_driver_tm_missing_data <- function() {
   data <- within(simple_teal_data(), {
+    set.seed(123)
     require(nestcolor)
 
     add_nas <- function(x) {


m7pr:
Yeah, so without `set.seed(123)` I get different data each time, hence different cut.off value

image

as this combination cutoff is dependent on the number of NAs in the data

```r
   combination_cutoff_q <- reactive({
      req(common_code_q())
      teal.code::eval_code(
        common_code_q(),
        quote(
          combination_cutoff <- ANL %>%
            dplyr::mutate_all(is.na) %>%
            dplyr::group_by_all() %>%
            dplyr::tally() %>%
            dplyr::ungroup()
        )
      )
    })

    output$cutoff <- renderUI({
      x <- combination_cutoff_q()[["combination_cutoff"]]$n

      # select 10-th from the top
      n <- length(x)
      idx <- max(1, n - 10)
      prev_value <- isolate(input$combination_cutoff)
      value <- `if`(
        is.null(prev_value) || prev_value > max(x) || prev_value < min(x),
        sort(x, partial = idx)[idx], prev_value
      )

      teal.widgets::optionalSliderInputValMinMax(
        ns("combination_cutoff"),
        "Combination cut-off",
        c(value, range(x))
      )
    })
```

m7pr:
And with `set.seed` before data creation, the data input is the same, hence the calculations for the output are the same
image

m7pr (approved):

PR #774, opened 2024-08-14
Ensures required packages are in Imports section

### LINKED ISSUE #518: Incorrect soft dependencies
We have moved some of the dependencies into Suggest but I think it was not a good call. As a result, when deploying an app that uses tmg modules, `rsconnect` is not able to identify app dependencies completely (it would not include package suggests). App developer needs to add explicit library calls into the app code but first of all he need to be aware of it (which is not very obvious). This is something we would like to avoid and keep the entry bar relatively low.

The trick about keeping dependencies in Suggests is not necessarily wrong but it has to be used only for **soft** dependencies so that the code is fully functional without a package. Great example can be found [here](https://github.com/insightsengineering/teal.code/blob/89f07c138f43d7fadaf8a03f8d46f211e3e6720d/R/qenv-eval_code.R#L96-L102). Here, we always calling `stop()` which indicates a hard dependency.

In order to fix it - let's move it back to be a hard dependency or make it a truly soft one.

As a direct consequence of this we have issues in `teal.gallery` and we had to add a lot of explicit library calls. Once fixed, please remove those.

averissimo:
# Pull Request

Fixes #518 

Note on removals:

- Remove `broom` as the `broom:glances` method can be replaced with `generics::glances`
  - According to my tests (methods are being deprecated as well)

#### Changes description

- Removed conditional use of packages
- Removes `{broom}` dependency
- Uses `rlang ` conditionally
- Uses Suggested packages conditional on tests


tests/testthat/test-shinytest2-tm_data_table.R
```diff
@@ -44,6 +44,7 @@ test_that("e2e - tm_data_table: Verify checkbox displayed over data table", {
 })
 
 test_that("e2e - tm_data_table: Verify module displays data table", {
+  testthat::skip_if_not_installed("rvest")
```
pawelru, line 47: I have recently pushed changes into the `teal` due to noSuggest check and the class definition looks like this:
https://github.com/insightsengineering/teal/blob/7eac4d58a5373aa511f363a2f259d850749501f7/R/TealAppDriver.R#L10-L20

Just thinking loud now, would it help if we would have the following instead?
```
    if (!requireNamespace("shinytest2", quietly = TRUE)) {
      if (requireNamespace("testthat", quietly = TRUE) && getNamespace("testthat")$is_testing()) {  
       getNamespace("testthat")$skip("rvest is not installed")
      } else {
        stop("Please install 'shinytest2' package to use this class.")
      }
    }
```

My motivation is to avoid a lot of skip statements in the upstream packages like here.

m7pr:
@averissimo is this still relevant? 
Should we add to this sprint and get someone to review?

averissimo:
@m7pr it is, although it needs a new pass to make sure that nothing new has created a conflict here

I'm moving it to draft

tests/testthat/test-shinytest2-tm_data_table.R
averissimo: That's awesome, I'll create the PR in {teal} to do just this. I think the `$is_testing()` is not required as `skip()` throws an error and it's on testthat to capture it and handle

DESCRIPTION
```diff
@@ -31,16 +31,30 @@ Depends:
     teal.transform (>= 0.5.0.9015)
```
averissimo, line 31: Teal requirement needs to be bumped with https://github.com/insightsengineering/teal/pull/1432

Possibly to `teal (>= 0.15.2.9091),`

Keeping conversation opened until the PR is closed

R/tm_missing_data.R
```diff
@@ -1135,6 +1127,13 @@ srv_missing_data <- function(id, data, reporter, filter_panel_api, dataname, par
         ggtheme = input$ggtheme
       )
 
+      # Unlikely that `rlang` is not available, new hashing may be expensive
+      hashing_function <- if (requireNamespace("rlang", quietly = TRUE)) {
+        quote(rlang::hash)
+      } else {
+        function(x) paste(as.integer(x), collapse = "")
```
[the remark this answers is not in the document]
averissimo: `rlang` was recently added to `Imports` as part of decorators, but it will be removed when the latest changes from `tmc` are integrated

DESCRIPTION
pawelru, line 1: I see changes in dependencies and I immediately think about verdepcheck. Can you please confirm that it is passing on this? If not - let's work on the deps versions.

This might need to wait for https://github.com/insightsengineering/teal.modules.general/pull/774/files#r1890465398

DESCRIPTION
llrs-roche: I approved that PR, so working on this PR can continue.

DESCRIPTION
```diff
@@ -31,16 +31,30 @@ Depends:
...
+    ggpp,
+    ggrepel,
+    goftest,
     grid,
+    gridExtra,
+    htmlwidgets,
+    jsonlite,
+    lattice (>= 0.18-4),
+    MASS,
+    rtables (>= 0.6.8),
```
llrs-roche, line 50: With all these additions this won't pass CRAN checks of at most 20 non base dependencies! See this issue for tmc: https://github.com/insightsengineering/coredev-tasks/issues/604
I know this goes against the purpose of this PR or the issue #518 but I think we need to keep it in mind. 

Can some of them be removed? For example, gridExtra already imports grid, so can grid be omitted from here (not sure if it can be done or not)?

llrs-roche (commented):
I left a couple of comments, as I haven't checked the code locally yet

DESCRIPTION
averissimo: That note from `CRAN` checks was the reason these packages were initially moved to `Suggests`. However, it's not a hard requirement.

With {tmc} we have ~27 non-base dependencies and some time ago we found packages in cran that have around the same number of deps or more.

If this is rejected we could try to argue the necessity, as these are real dependencies that are being used and have an effect on its usage (see library calls in teal.gallery)

We will be able to remove `lifecycle` when integrating the new changes from `tmc`, but others are very hard to replace.

The alternative is make them truly soft-dependencies

llrs-roche: > That note from `CRAN` checks was the reason these packages were initially moved to `Suggests`.
> However, it's not a hard requirement.
I thought it was that reason that made you move to Suggests. 

The requirement is not very strict for now...
I guess time will tell

teal.modules.general.Rproj
```diff
@@ -1,4 +1,5 @@
 Version: 1.0
+ProjectId: 003390c8-e31f-4b87-957d-7943778020cc
```
llrs-roche: New Rstudio version :) I wonder what this is used for, as it was not disclosed on the release notes

tests/testthat/test-shinytest2-tm_data_table.R
pawelru: just for the record: https://github.com/insightsengineering/teal/pull/1432

closing

teal.modules.general.Rproj
m7pr: I would opt for removing this, if this is different on different platforms or machines and we might keep over-writing this

llrs-roche: @m7pr I think once it is present it isn't rewritten, but it is true that is different for different machines or setups, mine was different for this PR (not sure how it is decided or how it will be used as it was not disclosed by Posit).

llrs-roche:
I thought this could affect the release, and the scheduled CI, so decided to pick it. I updated the PR and fixed a note about lifecycle.

```
Namespace in Imports field not imported from: 'lifecycle'
    All declared Imports should be used.
```

Feel free to review it @m7pr and decide how to proceed.

m7pr:
Thanks @llrs-roche 
Would you be able to start a Scheduled job on this PR?

m7pr:
Started the job in here https://github.com/insightsengineering/teal.modules.general/actions/runs/12787826187

m7pr:
needed to vbump teal.reporter, restarting the job

m7pr:
https://github.com/insightsengineering/teal.modules.general/actions/runs/12789415938

m7pr:
Scheduled job is still failing because of the inability to install teal.slice.
Updated teal version to overcome this.
Restarted job in here https://github.com/insightsengineering/teal.modules.general/actions/runs/12806234604

DESCRIPTION
m7pr: Yes, once min_cohort is satisfied I will try to put minimal versions of packages in the DESCRIPTION file

m7pr:
Needed to extend extra-deps parameter of the job configuration (Scheduled yaml in .github/workflows folder) so that min version of teal.slice is installed, but the DESCRIPTION file is not polluted as this package is not really a direct dependency of tmg

New job https://github.com/insightsengineering/teal.modules.general/actions/runs/12807617520

m7pr:
Also specified version of SummarizedExperiment in extra-deps for teal.slice. So a 3rd level dependency.
Restarting the build https://github.com/insightsengineering/teal.modules.general/actions/runs/12807968854

m7pr:
Also added MultiAssayExperiment to extra-deps, to get teal.slice to install.
New job https://github.com/insightsengineering/teal.modules.general/actions/runs/12808942137

m7pr:
One more https://github.com/insightsengineering/teal.modules.general/actions/runs/12826038718

m7pr (approved):

PR #799, opened 2024-11-19
introduce decorators for `tm_data_table`

m7pr:
Part of https://github.com/insightsengineering/teal/issues/1370

<details><summary> Working Example </summary>

```r


devtools::load_all("../teal")
devtools::load_all(".")

# general data example
data <- teal_data()
data <- within(data, {
  require(nestcolor)
  iris <- iris
})

custom_table_decorator_interactive <- teal_transform_module(
  ui = function(id) {
    ns <- NS(id)
    div(
      selectInput(
        ns("style"), 
        "Table Style", 
        choices = c("Default", "Striped", "Hover"), 
        selected = "Default"
      )
    )
  },
  server = function(id, data) {
      reactive({
        req(data(), input$style)
        within(data(), 
               {
                 if (style == "Striped") {
                   table <-
                     DT::formatStyle(
                       table,
                       columns = attr(table$x, "colnames")[-1],
                       target = 'row',
                       backgroundColor = '#f9f9f9'
                     )
                 } else if (style == "Hover") {
                   table <-
                     DT::formatStyle(
                       table,
                       columns = attr(table$x, "colnames")[-1],
                       target = 'row',
                       backgroundColor = '#f0f0f0'
                     )
                 }
               }, 
               style = input$style
        )
      })
    })
  }
)

app <- init(
  data = data,
  modules = modules(
    tm_data_table(
      variables_selected = list(
        iris = c("Sepal.Length", "Sepal.Width", "Petal.Length", "Petal.Width", "Species")
      ),
      dt_args = list(caption = "IRIS Table Caption"),
      decorators = list(custom_table_decorator_interactive)
    )
  )
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```

</details>


m7pr:
Hey @averissimo this PR is done, however I would need some extra pair of 👁️ eyes.

This example for for non-ineractive decorator. However it fails for interactive decorator because inside the `within` function, it can not find `input$style`. It is empty. So I think I put UI part in the wrong place inside the code. 

Would you maybe be able to take a look?

R/tm_data_table.R
```diff
@@ -282,6 +299,7 @@ ui_data_table <- function(id,
   tagList(
     teal.widgets::get_dt_rows(ns("data_table"), ns("dt_rows")),
     fluidRow(
+      ui_teal_transform_data(ns("decorate"), transformators = decorators),
```
averissimo, line 302: @m7pr the UI had the wrong id and was in the `ui_page_data_table` which created a different namespace.

Moving it to the "dataname" tab fixes the problem and allows each one to have their own decorator.

R/tm_data_table.R
```diff
@@ -305,36 +323,65 @@ srv_data_table <- function(id,
...
+                           server_rendering,
+                           decorators) {
   moduleServer(id, function(input, output, session) {
     iv <- shinyvalidate::InputValidator$new()
     iv$add_rule("variables", shinyvalidate::sv_required("Please select valid variable names"))
     iv$add_rule("variables", shinyvalidate::sv_in_set(
       set = names(isolate(data())[[dataname]]), message_fmt = "Not all selected variables exist in the data"
     ))
     iv$enable()
+    # teal::validate_inputs(iv)
```
averissimo, line 335: This won't pass `linter`, can you remove it?

R/tm_data_table.R
```diff
@@ -305,36 +323,65 @@ srv_data_table <- function(id,
...
+      req(data_table_data())
+      # no table is displayed
+      decorated_data_table_data()[["table"]]
     })
+
+    teal.widgets::verbatim_popup_srv(
+      id = "rcode",
+      verbatim_content = reactive(teal.code::get_code(req(decorated_data_table_data()))),
+      title = "R Code for PCA"
+    )
```
averissimo, line 385: "Show R Code is not working for me, can you confirm it?

Do we need to have the button? it wasn't there before

R/tm_data_table.R
```diff
@@ -29,6 +29,14 @@
 #'
 #' @inherit shared_params return
 #'
+#' @section Decorating `tm_data_table`:
+#'
+#' This module generates the following objects, which can be modified in place using decorators:
+#' - `table` (`DT::datatable`)
```
averissimo, line 35: (optional): what do you think about using a link?

```suggestion
#' - `table` ([DT::datatable()])
```

averissimo (changes requested):
Show R code is not working and there will be a linter error

R/tm_data_table.R
m7pr: I think I need to uncomment, this is a remaining from testing phase

R/tm_data_table.R
m7pr: @averissimo the Show R Code was in the wrong UI place and there was a namespace collision. Just moved it https://github.com/insightsengineering/teal.modules.general/pull/799/commits/d769821f66c71bb33365a78aec1884fff5ee5c4f

But the code looks super weird. I think we should put if statements around `eval_code` and not inside `eval_code`

image

m7pr: Check out all those weird `if(FALSE)` statements or `if(!is.null(NULL)`.
But I'm not sure it's worth it.

The decorator code, for example would need to look like this

```r
      reactive({
        req(data())
        
        if (input$style == "Striped") {
          within(data(), 
           {
             
               table <-
                 DT::formatStyle(
                   table,
                   columns = attr(table$x, "colnames")[-1],
                   target = 'row',
                   backgroundColor = '#f9f9f9'
                 )
             
           }
          )
        }  else if (input$style == "Hover") {
          within(data(), 
                 {
                  table <-
                    DT::formatStyle(
                      table,
                      columns = attr(table$x, "colnames")[-1],
                      target = 'row',
                      backgroundColor = '#f0f0f0'
                    )
                 }
          )
        } else {
          data()
        }
          
      })
```

instead of

```r
      reactive({
        req(data(), input$style)
        within(data(), 
               {
                 if (style == "Striped") {
                   table <-
                     DT::formatStyle(
                       table,
                       columns = attr(table$x, "colnames")[-1],
                       target = 'row',
                       backgroundColor = '#f9f9f9'
                     )
                 } else if (style == "Hover") {
                   table <-
                     DT::formatStyle(
                       table,
                       columns = attr(table$x, "colnames")[-1],
                       target = 'row',
                       backgroundColor = '#f0f0f0'
                     )
                 }
               }, 
               style = input$style
        )
      })
```

R/tm_data_table.R
m7pr: @averissimo when I uncomment this line the module does not produce an output : D

averissimo:
The \*\***CURRENT**\*\* implementation of data-table has for each of the datanames:

- 1 call of decorator UI/SRV
    - So each may have a different UI with options specific to each tab
 
### Q: Do we want to try and create a "singleton" UI that then applies to each one dataname?

In practice: Do we want to have a single UI for options applied to each dataname, or a UI for each of the datanames (as is now)

Consequences: We need get inputs from 1-level above as the `srv_teal_transform_data` is called from within the `dataname` prefixed namespace (`data_table-{iris,mtcars, ...}`)

---

#### Current implementation vs. Q

image

R/tm_data_table.R
averissimo: I don't think we should add new functionality to the module on the PR.

If you are feeling conflicted open a new issue.

R/tm_data_table.R
averissimo: Solved in f94a6586

PR #818, opened 2024-12-13
87 remove datasets - decrease package size

m7pr:
In the spirit of package size reduction https://github.com/insightsengineering/nestdevs-tasks/issues/87
I replaced all the calls for tmg datasets with their respective datasets from teal.data.
This way we no longer need to have datasets copies in tmg.


llrs-roche (approved):
Looks great. LGTM


While checking the package locally I got an unrelated error from running a vignette:
"there is no package called ggpmics" there is a typo there and should be ggpmisc. Also on that vignette using-scatterplot.Rmd, the [suggested dependency checks](https://github.com/insightsengineering/teal.modules.general/blob/46e5267951f341d7245042a1826d31406c0f2417/vignettes/using-scatterplot.Rmd#L24) are not silent: `requireNamespace("ggpmics")` should be `requireNamespace("ggpmics", quietly = TRUE)` to avoid raising an error if they are not present. I'll open an issue to not forget about these.

m7pr:
Thanks @llrs-roche I pushed those 2 changes that you mentioned

llrs-roche:
Oh, I had opened a different issue to avoid doing too much here: https://github.com/insightsengineering/teal.modules.general/issues/819

But if you do close that, be aware that the `quietly = TRUE` argument should be on all the `requireNamespace` calls, not only ggmics . Otherwise CRAN might come with a request to fix the package at any time.  

Unrelated to this problem there seems to be an issue with `join_keys<-`, but I am not sure why this PR triggers that. Probably to fix it we might need to raise the dependency version required of on teal.data .

m7pr:
Got it! Will continue on Monday!

PR #825, opened 2025-01-13
Use the app modifiers in favor of `init` args

vedhav:
Part of https://github.com/insightsengineering/teal/issues/1310


m7pr (approved):

R/tm_front_page.R
```diff
@@ -53,11 +53,12 @@
...
 #'       show_metadata = TRUE
 #'     )
-#'   ),
-#'   header = tags$h1("Sample Application"),
-#'   footer = tags$p("Application footer"),
+#'   )
 #' )
 #'
+#' app <- modify_header(app, tags$h1("Sample Application"))
+#' app <- modify_footer(app, tags$p("Application footer"))
```
gogonzo, line 60: We can use the `|>` in the documentation as related `teal` PR depends on 4.1 
```suggestion
#'   )
#' ) |>
#'   modify_header(tags$h1("Sample Application")) |>
#'   modify_footer(tags$p("Application footer"))
```

vedhav: 👍🏽 also bumped the R version to `4.1` over here.

PR #828, opened 2025-01-21
add a check for non-unique decorator names

m7pr:
The same as in tmc
https://github.com/insightsengineering/teal.modules.clinical/pull/1313/commits/b048f3744c475e3a2e19733d0b34d71061548faf
do not allow decorators with duplicated names


llrs-roche (changes requested):
If there are duplicated decorators and decorators that won't be used by the module only the first issue will be raised.

m7pr:
@llrs-roche Just sent the same fix as in here

llrs-roche (approved):

PR #839, opened 2025-02-03
Add missing transformators arg

gogonzo:
`transformators` arg is addressed in `teal`. Every module which has `data` in the server needs a `transformators` argument


m7pr (approved):
@gogonzo Those were omitted during decorators phase, that's why we didn't add any transformators. But I guess transformator has more sense than decorator in here, as we can edit input data with transformators. Decorators do not have much sense in here as there isn't any output that can be really edited.

PR #840, opened 2025-02-03
823 block test table without validation error

### LINKED ISSUE #823: `tm_g_distribution` always starts with validation error
### What happened?

This is not necessarily a bug but is does interfere in some cases.

`tm_g_distribution` starts with `Tests`* (`id = 'teal-main_ui-root-dist_1-module-dist_tests_input'`) unselected, which produces a validation error in the Tests output. This results in a validation error. Since it is impossible to set the theoretical distribution in the module call, this is inevitable.

Would it be possible to either 
a) blocking the Tests output be achieved without raising a validation error,
b) add an argument to `tm_g_distribution`?

(*) probably should be "Test", by the way

### sessionInfo()

_No response_

### Relevant log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

chlebowa:
Fixes #823


m7pr:
Main

image


Branch

image

m7pr (approved):

PR #862, opened 2025-02-24
[WIP] teal.reporter redesign poc

m7pr:
Companion to
- [teal.reporter/redesign@main](https://github.com/insightsengineering/teal.reporter/tree/redesign%40main)
- [teal/test@bslib@main](https://github.com/insightsengineering/teal/tree/test%40bslib%40main)

<details><summary> code </summary>

```r

devtools::load_all('../teal.reporter')
devtools::load_all('../teal')
devtools::load_all('.')

# general data example
data <- teal_data()
data <- within(data, {
  require(nestcolor)
  CO2 <- CO2
})

app <- init(
  data = data,
  modules = modules(
    tm_a_regression(
      label = "Regression",
      response = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variable:",
          choices = "uptake",
          selected = "uptake",
          multiple = FALSE,
          fixed = TRUE
        )
      ),
      regressor = data_extract_spec(
        dataname = "CO2",
        select = select_spec(
          label = "Select variables:",
          choices = variable_choices(data[["CO2"]], c("conc", "Treatment")),
          selected = "conc",
          multiple = TRUE,
          fixed = FALSE
        )
      )
    )
  )
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```

</details>

image

### DIFF
--- a/R/tm_a_regression.R
+++ b/R/tm_a_regression.R
@@ -282,9 +282,6 @@ ui_a_regression <- function(id, ...) {
       tags$div(verbatimTextOutput(ns("text")))
     )),
     encoding = tags$div(
-      ### Reporter
-      teal.reporter::simple_reporter_ui(ns("simple_reporter")),
-      ###
       tags$label("Encodings", class = "text-primary"),
       teal.transform::datanames_input(args[c("response", "regressor")]),
       teal.transform::data_extract_ui(
@@ -384,7 +381,6 @@ ui_a_regression <- function(id, ...) {
 # Server function for the regression module
 srv_a_regression <- function(id,
                              data,
-                             reporter,
                              filter_panel_api,
                              response,
                              regressor,
@@ -393,7 +389,6 @@ srv_a_regression <- function(id,
                              ggplot2_args,
                              default_outlier_label,
                              decorators) {
-  with_reporter <- !missing(reporter) && inherits(reporter, "Reporter")
   with_filter <- !missing(filter_panel_api) && inherits(filter_panel_api, "FilterPanelAPI")
   checkmate::assert_class(data, "reactive")
   checkmate::assert_class(isolate(data()), "teal_data")
@@ -1034,25 +1029,20 @@ srv_a_regression <- function(id,
     )
 
     ### REPORTER
-    if (with_reporter) {
-      card_fun <- function(comment, label) {
-        card <- teal::report_card_template(
-          title = "Linear Regression Plot",
-          label = label,
-          with_filter = with_filter,
-          filter_panel_api = filter_panel_api
-        )
-        card$append_text("Plot", "header3")
-        card$append_plot(plot_r(), dim = pws$dim())
-        if (!comment == "") {
-          card$append_text("Comment", "header3")
-          card$append_text(comment)
-        }
-        card$append_src(source_code_r())
-        card
+    card_fun <- reactive({
+      req(plot_r(), source_code_r())
+        function() {
+          card <- teal.reporter::report_document()
+          card <- c(card, c(list("## Plot")))
+          # card <- c(card, c(list("## Plot"), list(plot_r())))
+          # card <- c(card, list(source_code_r()))
+          card
       }
-      teal.reporter::simple_reporter_srv("simple_reporter", reporter = reporter, card_fun = card_fun)
-    }
+    })
     ###
+
+    list(
+      report_card = card_fun
+    )
   })
 }
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -79,6 +79,9 @@ Suggests:
     shinytest2,
     testthat (>= 3.1.9),
     withr (>= 2.0.0)
+Remotes:
+    insightsengineering/teal@test@bslib@main,
+    insightsengineering/teal.reporter@redesign_poc@main
 VignetteBuilder:
     knitr,
     rmarkdown
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -81,7 +81,7 @@ Suggests:
     withr (>= 2.0.0)
 Remotes:
     insightsengineering/teal@test@bslib@main,
-    insightsengineering/teal.reporter@redesign_poc@main
+    insightsengineering/teal.reporter@redesign@main
 VignetteBuilder:
     knitr,
     rmarkdown


R/tm_a_regression.R
```diff
@@ -384,16 +381,15 @@ ui_a_regression <- function(id, ...) {
 # Server function for the regression module
 srv_a_regression <- function(id,
                              data,
-                             reporter,
                              filter_panel_api,
                              response,
                              regressor,
+                             reporter,
```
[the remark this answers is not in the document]
m7pr: I think it's fine if it stays

R/tm_a_regression.R
```diff
@@ -1037,25 +1033,52 @@ srv_a_regression <- function(id,
...
+  # base64enc::base64encode(tmpfile)
+  base64_string<- knitr::image_uri(tmpfile)
+  sprintf("![Plot](%s)", base64_string)
+}
+#
+# plot_to_base64(qqplot(x = 1:2, y = 1:2))
+
+format_as_markdown_chunk <- function(r_code) {
+  # Wrap the code inside a Markdown R chunk
+  sprintf("```{r}\n%s\n```", r_code)
```
m7pr, line 1074:
```suggestion
  sprintf("```{r, eval=FALSE}\n%s\n```", r_code)
```

R/tm_a_regression.R
```diff
@@ -1037,25 +1033,26 @@ srv_a_regression <- function(id,
...
-      teal.reporter::simple_reporter_srv("simple_reporter", reporter = reporter, card_fun = card_fun)
-    }
+    card_fun <- reactive({
+      req(plot_r(), source_code_r())
+      teal.reporter::report_document(
+
+        "## Plot",
+        plot_r(),
+        "## Source Code",
+        format_as_markdown_chunk(source_code_r()),
```
gogonzo, line 1043: There is no such thing in markdown syntax as "chunk", this term is rmarkdown specific. So the word markdown and chunk can't go along. I suggest `as.code_chunk` or [`as.code_block`](https://www.markdownguide.org/basic-syntax/#code-blocks-1). This function should be in `teal.reporter` because this will be necessary in all teal.module.*

```suggestion
        teal.reporter::as.code_block(source_code_r()),
```

R/tm_a_regression.R
gogonzo: I think eval=FALSE or any other chunk parameter shouldn't be given by default. Parameters should be arguments of the `as.code_block` function. I'm not sure at this point whether "code_block_elem" should be kept in the list as-is with the special class or if the backticks-enclosure should be added immediately.

R/tm_a_regression.R
```diff
@@ -1036,25 +1066,58 @@ srv_a_regression <- function(id,
...
+          header = LETTERS[1:3],
+          rtables::rrow("one to three", 1, 2, 3),
+          rtables::rrow("more stuff", rtables::rcell(pi, format = "xx.xx"), "test", "and more")
+        ),
+
+        "## Table for testing",
+        teal.reporter::code_chunk(
+          "head(iris)"
+        ),
+        head(iris)
```
m7pr, line 1114: WORD format does not yet works for data.frames. will try to fix that today

R/tm_a_pca.R
```diff
@@ -1132,30 +1183,43 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
+
+        "### Eigenvectors Table",
+        teal.reporter::code_chunk(computation_tbl_eig_code_r()),
+        computation()[["tbl_eigenvector"]],
+
+        "### Plot",
+        teal.reporter::code_chunk(
+          plot_code_r() |> styler::style_text() |> paste(collapse = "\n")
+        ),
+        plot_r()
```
gogonzo, line 1216: outputs added to the reportDocument are independent of the code chunk here. It means they are not linked to particular chunk. Implementation in teal.reporter removes all outputs with no exceptions. What if module developer wants to add a plot as an image/table without any code chunk? 

```suggestion
        teal.reporter::code_chunk(
          plot_code_r() |> styler::style_text() |> paste(collapse = "\n"),
          echo = plot_r()
        )
```

m7pr: Yes, at some point I had a linking function but then realized it doesn't give much value.
If you don't want to add any code chunk, then just do not include code in `report_document()`.
If you want to keep some outputs that are R objects, let's add attribute that states they should be kept.

So here are some scenarios below:


1. Code in the report and output but plot only in the output.

```r
teal.reporter::report_document(
  teal.reporter::code_chunk(plot_code),
  plot_object
)
```

2. Code in the report and output and plot in the report and output.

```r
keep_in_report <- function(x, keep = TRUE) {
    attr(x, "keep_in_report") <- keep
    x
}
teal.reporter::report_document(
  teal.reporter::code_chunk(plot_code),
  plot_object |> keep_in_report()
)
```

3. No code in the report and no code in the output and plot in the report and output.

```r
keep_in_report <- function(x, keep = TRUE) {
    attr(x, "keep_in_report") <- keep
    x
}
teal.reporter::report_document(
  plot_object |> keep_in_report()
)
```

4. No code in the report and no code in the output and plot only in the output.

```r
teal.reporter::report_document(
  plot_object
)
```

And cases when you want to have code in the output, but not in the report

```r
keep_in_report <- function(x, keep = TRUE) {
    attr(x, "keep_in_report") <- keep
    x
}
teal.reporter::report_document(
  teal.reporter::code_chunk(plot_code) |> keep_in_report(keep = FALSE),
  plot_object |> keep_in_report()
)
```

m7pr: Also, is you don't want to include code, but just outputs, then this breaks the reproducibility principle. I think we should motivate and influence developers to include as much code as possible, so everything can be reproduced.

And if they don't want to keep the code, `keep_in_report(keep = FALSE)` could do the trick, or you just don't put the code in `report_document()` - it is not evaluated anyway.

gogonzo: Nice alternative. Please add it to the teal.reporter. We can discuss this later

R/tm_a_pca.R
```diff
@@ -1124,6 +1123,58 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
     })
 
     # Render R code.
+    subset_code <- function(code, data) {
+      gsub(code, "", teal.data::get_code(data), fixed = TRUE)
+    }
```
gogonzo, line 1128: This is why I've created an issue in teal.code some time ago:

https://github.com/insightsengineering/teal.code/issues/241

m7pr: Yes, was thinking the exact same thing few days ago : ) This would be so helpful in here.
I think I could work some proposal for teal.code

m7pr: Hey, thanks to https://github.com/insightsengineering/teal.code/pull/248 I allowed code to be labeled, like in this commit c1abab492f4d414d150baf4caececaf3f08ab0cc

m7pr: The only drawback of this approach is that transformator/decorator needs to use the same `label = ` value in it's `eval_code/within` calls, to get the code to be included in the final report.

R/tm_a_pca.R
m7pr: Alrighty, will try to wrap something for this tomorrow : ) (Friday)

R/tm_a_regression.R
```diff
@@ -1026,7 +1033,41 @@ srv_a_regression <- function(id,
...
     })
 
-    # Render R code.
+    # # Render R code.
+    # subset_code <- function(code, data) {
+    #   gsub(code, "", teal.data::get_code(data), fixed = TRUE)
+    # }
+
+    setup_code_r <- reactive(teal.data::get_code(req(data())))
+    libraries_code_r <- reactive(teal.code::get_code(req(decorated_output_q()), labels = "libraries"))
```
gogonzo, line 1042: Makes sense. Let's not focus on the code-splitting and finish the fundamental part of teal.reporter. We can come back to this discussion at some point

R/tm_a_pca.R
m7pr: Also an example on how to remove things from report but keep in output https://github.com/insightsengineering/teal.reporter/pull/307/commits/707b6bd0631ef134dec150f2a8a4eb59f2763d5e

m7pr: Note: previewer shows all objects, no matter what's their `keep` status.

R/tm_a_regression.R
```diff
@@ -1036,25 +1048,71 @@ srv_a_regression <- function(id,
...
+        teal.reporter::code_chunk(
+          "head(iris)"
+        ),
+        head(iris),
+
+        "## keep_in_report",
+
+        "If you don't want to include code for head(mtcars) in report - just don't include it.",
+
+        "If you want the object to be kept in the report as loaded from .rds use keep_in_report()",
```
gogonzo, line 1102:
```suggestion
        "If you want the object to be kept in the rendered output (pdf, pptx, etc.) use keep_in_report()",
```

For me high-level explaination is. If `head(mtcars) |> keep_in_report(FALSE)`:
- it will be included in the Reporter object (because it is in the list of items provided to report_document) 
- will be shown when previewing the results in the report_previewer (to avoid double evaluation of the code chunk)
- will be included in rendered in pdf, pptx, ... files directly to avoid double evaluation of the code chunk
- Will be excluded from saved .Rmd file content because .Rmd doesn't need to include a rendered output - output will be rendered when runnning .Rmd file

If `keep_in_report(TRUE)` - then always included

m7pr: Yes, your assumptions are right. I wonder if that's gonna be easy to grasp for the final user 🤔

R/tm_a_pca.R
```diff
@@ -354,7 +351,6 @@ ui_a_pca <- function(id, ...) {
 
 # Server function for the PCA module
 srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, plot_width, ggplot2_args, decorators) {
-  with_reporter <- !missing(reporter) && inherits(reporter, "Reporter")
```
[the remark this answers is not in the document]
m7pr: 2f2590656cf458c9b7953633f6675b6ff0703938

R/tm_a_pca.R
```diff
@@ -1124,38 +1140,57 @@ srv_a_pca <- function(id, data, reporter, filter_panel_api, dat, plot_height, pl
...
+        "## PCA Model",
+        teal.reporter::code_chunk(computation_model_code_r()),
+
+        "### Principal Components Table",
+        teal.reporter::code_chunk(computation_tbl_imp_code_r()),
+        computation()[["tbl_importance"]],
+
+        "### Eigenvectors Table",
+        teal.reporter::code_chunk(computation_tbl_eig_code_r()),
+        computation()[["tbl_eigenvector"]],
```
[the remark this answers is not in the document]
m7pr: 2aaadcbb6796f37c663b547b9a5e17629626b084

R/tm_a_pca.R
m7pr: continuation of this thread is here https://github.com/insightsengineering/teal.modules.general/pull/862#discussion_r2057591555

gogonzo:
@m7pr closing this one as it is not in use anymore 👍 Thanks

PR #867, opened 2025-02-27
add sparkline::sparkline to helper-TealAppDriver.R

m7pr:
Just a test that tries to solve https://github.com/insightsengineering/coredev-tasks/issues/624


donyunardi:
Let's not merge until I pushed tmg to CRAN.
The failures are for shinytest2, which are not blockers for CRAN release.

donyunardi:
I don't think we need to proceed with this PR. 
As I noted [here](https://github.com/insightsengineering/coredev-tasks/issues/624#issuecomment-2706917845), there seems to be issue with parallel computing.

PR #886, opened 2025-06-10
Remove dev_suppress

gogonzo:
Related to https://github.com/insightsengineering/teal.osprey/issues/302


m7pr:
R: 4.4.0
Editor: Cursor (extension of visual studio code): 1.0.0 / based on VSCode Version 1.96.2

tm_a_regression shows the text output

image

tm_g_association shows the title

image

However, once those modules are called, and I close the app, the first next call in the console opens the last plot from the app

image

m7pr:
I think `teal.code::dev_suppress` is needed so that plots are not shown in the IDE

PR #891, opened 2025-07-04
Don't update distribution params labels when variable is deselected in `tm_g_distribution`

### LINKED ISSUE #510: labels in distribution module resetting
If you remove the variable:
image

You lose these labels:
image

From https://github.com/insightsengineering/teal.modules.general/pull/509

m7pr:
Fixes #510

### DIFF
--- a/R/tm_g_distribution.R
+++ b/R/tm_g_distribution.R
@@ -569,50 +569,103 @@ srv_distribution <- function(id,
       }
     })
 
+    params_manually_set <- reactiveVal(FALSE)
+    prev_t_dist <- reactiveVal(NULL)
+    prev_variable <- reactiveVal(NULL)
+
+    get_dist_params <- function(x, dist) {
+      if (dist == "unif") {
+        return(stats::setNames(range(x, na.rm = TRUE), c("min", "max")))
+      }
+      tryCatch(
+        MASS::fitdistr(x, densfun = dist)$estimate,
+        error = function(e) c(param1 = NA_real_, param2 = NA_real_)
+      )
+    }
+
     observeEvent(
       eventExpr = list(
         input$t_dist,
-        input$params_reset,
         selector_list()$dist_i()$select
       ),
       handlerExpr = {
-        params <-
-          if (length(input$t_dist) != 0) {
-            get_dist_params <- function(x, dist) {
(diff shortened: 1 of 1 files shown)


donyunardi:
Originally, issue arises when user deselect the "Variable" selection. This triggers the label being resetted to param1 and param2, even though distribution is selected, in this case "normal".
Kapture 2025-07-07 at 11 23 46

When I use `main` version, I can't deselect the "Variable" selection as it always select a variable. 
Kapture 2025-07-07 at 11 33 30

Then only way to test this PR by setting the `select_spect(selected = NULL)` for the `dist_var` argument. However, I still see the same behavior where the labels are still showing 'param1' and 'param2', even though the distribution was selected.
Kapture 2025-07-07 at 11 35 41

R/tm_g_distribution.R
```diff
@@ -595,7 +595,17 @@ srv_distribution <- function(id,
...
+          map_distr_nams <- list(
+            normal = c("mean", "sd"),
+            lognormal = c("meanlog", "sdlog"),
+            gamma = c("shape", "rate"),
+            unif = c("min", "max")
+          )
+          params_names <- map_distr_nams[[input$t_dist]] %||% names(params)
+        } else {
+          params_names <- names(params)
+        }
```
donyunardi, line 608:
```suggestion
        map_distr_nams <- list(
          normal = c("mean", "sd"),
          lognormal = c("meanlog", "sdlog"),
          gamma = c("shape", "rate"),
          unif = c("min", "max")
        )
        params_names <- map_distr_nams[[input$t_dist]] %||% names(params)
```

Since you already using the `%||%` operator maybe we don't need the `if..else..` statement?
I tested this locally and seems to be working as expected.

donyunardi (changes requested):

R/tm_g_distribution.R
m7pr: I wanted to prevent `input$t_dist` being NULL, but will double check how it looks without the `if`

```r
> x <- list(a = 1, b = 2)
> x[[NULL]]
Error in x[[NULL]] : attempt to select less than one element in get1index
```

donyunardi (approved):

PR #905, opened 2025-07-30
Final UI/UX changes

vedhav:
Minor css cleanup


R/tm_g_distribution.R
```diff
@@ -260,7 +260,6 @@ ui_distribution <- function(id, ...) {
     encoding = tags$div(
       ### Reporter
       teal.reporter::add_card_button_ui(ns("add_reporter"), label = "Add Report Card"),
-      tags$br(), tags$br(),
```
averissimo, line 263: Is this enough space? It seems a bit crammed

image

Old

image

R/tm_g_scatterplot.R
```diff
@@ -362,18 +362,16 @@ ui_g_scatterplot <- function(id, ...) {
   )
 
   tagList(
-    include_css_files("custom"),
     teal.widgets::standard_layout(
       output = teal.widgets::white_small_well(
         teal.widgets::plot_with_settings_ui(id = ns("scatter_plot")),
-        tags$h1(tags$strong("Selected points:"), class = "text-center font-150p"),
+        tags$h1(tags$strong("Selected points:"), style = "text-align: center !important; font-size: 150%;"),
```
averissimo, line 368: Can we avoid the "!important"?

Is it because the bootstrap CSS has this as well?

vedhav: On the other hand I think the center align does not look that great, I'll remove it completely.

### Before
Screenshot 2025-08-01 055600

### Now
Screenshot 2025-08-01 055611

R/tm_g_distribution.R
vedhav: I'll revert this change. I think we still need the double line breaks. In `teal.reporter` I made a change to make this a `button` instead of `a` link which broke them in next line by default so I though just a break would be enough, but looks like having this space is better indeed.

averissimo (approved):

PR #907, opened 2025-08-11
Update lifecycle badges

m7pr:
Part of https://github.com/insightsengineering/coredev-tasks/issues/649

Changed deprecate soft -> deprecate hard for badges introduced in 0.4.0. Next release is 0.4.2 or higher.


gogonzo (approved):
👍

PR #942, opened 2025-10-30
picks

### LINKED ISSUE #36: Design data extract and data merge
# Motivation
Data extract/data merge has been criticised by many parties mainly for opaqueness and complicated API. The main goal of the refactor should be:
1. Remove independent filtering on the selector level, to avoid scenarios like below where ADLB is merged to ADLB. No other data-visualization tools offers such a functionality. Possible data transformation (reshape or filter) should be done once for dataset before variables selection. Independent filtering also complicates API call to initialise encoding panel. `data_extract_spec` sometimes covers more than hanf of the code space:

    ```r
    a <- data_extract_spect("ADSL", select_spec(...))
    b <- data_extract_spec("ADLB", filter_spec(<PARAMCD>), filter_spec(<AVISIT>), select_spec(...))
    c <- data_extract_spec("ADLB", select_spec(...))
    ``` 

    Screenshot 2022-05-18 at 06 37 53

    Selecting variables to the module should be minimalized to specifying variables only (without data transformation). Still open question whether we allow to select from different datasets (support automatic merge) or allow to select from one dataset only (no automatic merge).
    ```r
    a <- select_variables(<select from numeric vars in ADSL>)
    b <- select_variables(<AVAL from ADLB>)
    c <- select_variables(<select from factor vars in ADSL>)
    ```

2. If we decide to keep offering automatic merge we need to provide UI element informing that there is merge and how it's performed (used keys, shape of the final ANL)

Solutions in the comments.

gogonzo:
Closes https://github.com/insightsengineering/NEST-roadmap/issues/36

Check also with:
- https://github.com/insightsengineering/teal.transform/pull/270
- https://github.com/insightsengineering/teal/pull/1642

In this PR, due to the fact that `data_extract_spec` is not 1:1 convertible to `picks`, I propose to have an S3 method in each `tm_` to switch between data_extract_spec/picks depending on the input type. Eventually, `data_extract_spec` variant will be removed completely after deprecation period.


llrs-roche:
S3 methods can only dispatch according to one argument. This PR duplicates entire modules for a dispatch on one argument. Isn't it better to have a conversion or a warning on each argument? User might end up mixing the new and the old approach and that could throw off the dispatch mechanism. 
Still this would require that some part of the module's code would need to change but I don't think it would be that much. Specially if the full release and removal of the old des object.

gogonzo:
> Isn't it better to have a conversion or a warning on each argument?

It is better, problem is that delayed-filter_spec (with `value_choices()`) are not convertible to `teal_transform_module` and thus we would have a breaking change. 

I provided this "brute-force" solution to make sure nothing breaks. It is possible to implement filter_spec -> teal_transform_module conversion but it will require significant design intrusion. `value_choices(subset = function(data))` where `data` is a data.frame - this will require delivering `data.frame` instead of `vector` from variables to values in order to support `value_choices`.

Instead of making a design compromise in `resolver` and `determine.values`, I would rather vote for changing `filter_spec` to support `subset = function(column)` instead.

llrs-roche:
> > Isn't it better to have a conversion or a warning on each argument?
> 
> It is better, problem is that delayed-filter_spec (with `value_choices()`) are not convertible to `teal_transform_module` and thus we would have a breaking change.

Yes, my idea is to raise an error when they are not convertible, and just a deprecation warning when they can be converted. 

> 
> I provided this "brute-force" solution to make sure nothing breaks. It is possible to implement filter_spec -> teal_transform_module conversion but it will require significant design intrusion. `value_choices(subset = function(data))` where `data` is a data.frame - this will require delivering `data.frame` instead of `vector` from variables to values in order to support `value_choices`.
> 
> Instead of making a design compromise in `resolver` and `determine.values`, I would rather vote for changing `filter_spec` to support `subset = function(column)` instead.

I am never sure of how to ease these transitions between code/designs. There are many tradeoffs and the most important information is hard to estimate: user current usage and user adoption of the future new design. How long we'll we need to keep `value_choices` and `filter_spec` working? Even if it is a breaking change it could be with teal.modules.* 1.0.0 release.

DESCRIPTION
```diff
@@ -48,6 +48,7 @@ Imports:
     lattice (>= 0.18-4),
     lifecycle (>= 0.2.0),
     MASS (>= 7.3-60),
+    rlang,
```
m7pr:
```suggestion
    rlang (>= 1.1.0),
```

DESCRIPTION
```diff
@@ -83,8 +82,8 @@ VignetteBuilder:
     knitr,
     rmarkdown
 Remotes:
-    insightsengineering/teal,
-    insightsengineering/teal.reporter
+    insightsengineering/teal@redesign_extraction@main,
+    insightsengineering/teal.transform@redesign_extraction@main
```
m7pr: IMHO `teal.reporter` is still needed in `Remotes`, as it's not released on CRAN?

```suggestion
    insightsengineering/teal.reporter
    insightsengineering/teal.transform@redesign_extraction@main
```

R/zzz.R
```diff
@@ -7,4 +7,8 @@
 ggplot_themes <- c("gray", "bw", "linedraw", "light", "dark", "minimal", "classic", "void")
 
 #' @importFrom lifecycle deprecated
+#' @importFrom rlang :=
```
m7pr: Why is that needed? Is it becuase of the usage in `tm_missing_data`?

llrs-roche:
Commenting here so we don't forget: [tm_gtsummary](https://insightsengineering.github.io/teal.modules.general/main/reference/tm_gtsummary.html) will need to be added to this PR.

averissimo (approved):
🥳 🔥

PR #943, opened 2025-11-05
Explore testServer on modules

llrs-roche:
This is a branch to explore how `shiny::testServer` could help test modules.
This exploration is part of the work done for https://github.com/insightsengineering/coredev-tasks/issues/680

Just exposing it here because it doesn't seem possible (at this moment) to avoid validation errors triggered by default as there is a recursive dependency where `selector_list()` depends on `selector_list()`, which prevents to overwrite it for it to go to get other output (on `output_q()` from the tested `tm_g_association()`).

### DIFF
--- a/tests/testthat/test-tm_g_association.R
+++ b/tests/testthat/test-tm_g_association.R
@@ -1,0 +1,62 @@
+test_that("tm_g_association fails initially", {
+
+  data <- within(teal_data(), {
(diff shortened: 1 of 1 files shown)


m7pr:
@llrs-roche here I added some tests that test server functions. Similar to your approach
https://github.com/insightsengineering/teal.modules.general/pull/959

PR #944, opened 2025-11-05
Adds `rmarkdown` module

averissimo:
# Pull Request

- Fixes https://github.com/insightsengineering/coredev-tasks/issues/675

Thanks to @StefanThoma's for working on the first prototype

### Changes description

- Adds rmarkdown module
- Renders the output to module UI and to Reporter
- Source code is not verified as it adds a comment on how to render it
- Adds download button to get the original Rmd file
  - Option in module to enable or disable this functionality
  - downloaded Rmd has module's data included (data pre-processing + filter)
- Uses `extra_transform` parameter to add "generic decorators" to the top of the UI
  - name is temporary
- [x] Ensure figures are displayed properly in reporter (PDF output)
- [x] Avoid encoding images to base64 (mayb keep it in raw or equivalent)

### To discuss

- Should we render the `.R` to get all the R code from the report and add it to the code of the `qenv`?
- [limitation] Use of `markdown` when rendering doesn't print title nor author
  - Rendering it on the page breaks `teal`
  - Wrapping in `iframe` doesn't extend full height and then creates strange reports
- Should we use `transformers` to add extra input / data manipulation
  - Or keep a parameters that allows to add transforms _(for example extra inputs)_ to the module by the app developer

(topics below added from conversation with @gogonzo )

- should `teal_report` contain markdown elements?
- should `teal_data` evaluate `rmarkdown::render()` and include this to `@code`
- should markdown be evaluated with `rmarkdown::render(envir = <qenv>)` instead of `params = list()`?
  - visibility concern here on developing the Rmd

### App (same 2nd one in example, but with `download = TRUE`)

```r
pkgload::load_all("../teal.modules.general")

data <- teal_data()
data <- within(data, {
  CO2 <- CO2
  CO2[["primary_key"]] <- seq_len(nrow(CO2))
})
join_keys(data) <- join_keys(join_key("CO2", "CO2", "primary_key"))

static_decorator <- teal_transform_module(
  label = "N Cols selector",
  ui = function(id) {
    ns <- NS(id)
    tags$div(
      numericInput(ns("n_rows"), "Show n rows", value = 5, min = 0, max = 200, step = 5)
    )
  },
  server = function(id, data) {
      reactive({
        req(data())
        within(data(), {
          n_rows <- n_rows_value
        }, n_rows_value = input$n_rows)
      })
    })
  }
)

app <- init(
  data = data,
  modules = modules(
    tm_rmarkdown(
      label = "RMarkdown Module",
      rmd_content = readLines(system.file(file.path("sample_files", "co2_example.Rmd"), package = "teal.modules.general")),
      allow_download = TRUE,
      extra_transform = list(static_decorator)
    )
  )
) |> shiny::runApp()
```

### DIFF
--- a/tests/testthat/test-tm_rmarkdown.R
+++ b/tests/testthat/test-tm_rmarkdown.R
@@ -51,7 +51,7 @@ testthat::describe("srv_rmarkdown", {
         extra_transform = list(), allow_download = TRUE
       ),
       expr = {
-        .noWs <- c("before", "after", "outside", "after-begin", "before-end")
+        .noWs <- c("before", "after", "outside", "after-begin", "before-end") # nolint: object_name_linter.
         testthat::expect_equal(
           gsub("\n", "", as.character(output$rmd_output$html)),
           as.character(


R/tm_rmarkdown.R
```diff
@@ -0,0 +1,267 @@
...
+}
+
+# Server function for the rmarkdown module
+srv_rmarkdown <- function(id, data, rmd_file, allow_download, extra_transform) {
+  checkmate::assert_class(data, "reactive")
+  checkmate::assert_class(isolate(data()), "teal_data")
+  moduleServer(id, function(input, output, session) {
+    if (allow_download) {
+      output$download_rmd <- downloadHandler(
+        filename = function() basename(rmd_file),
```
gogonzo: shouldn't downloadable rmarkdown contain also "reproducible" qenv code? IMO `tm_rmarkdown(rmd_file)` is just a template and returning a template doesn't make sense if it lacks data to actually execute templete-file.

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,267 @@
...
+      "extra_transform",
+      data = pre_decorated_q_r,
+      transformators = extra_transform
+    )
+
+    rendered_path_r <- reactive({
+      datasets <- req(q_r()) # Ensure data is available
+      temp_dir <- tempdir()
+      temp_rmd <- tempfile(tmpdir = temp_dir, fileext = ".Rmd")
+      temp_html <- tempfile(tmpdir = temp_dir, fileext = ".md")
```
gogonzo: `temp_html` but `fileext = ".md"`

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,267 @@
...
+      tryCatch(
+        {
+          rmarkdown::render(
+            temp_rmd,
+            output_format = rmarkdown::md_document(
+              variant = "gfm",
+              toc = TRUE,
+              preserve_yaml = TRUE
+            ),
+            output_file = temp_html,
```
gogonzo: Probably `output_file` should remain empty as it will be given automatically - depending on the `output:` in yaml header. It could be `md/html/pdf` etc.
```suggestion
```

averissimo: Good point, `render()` returns the path

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,218 @@
...
+      )
+    })
+
+    rendered_path_r <- reactive({
+      datasets <- req(q_r())  # Ensure data is available
+
+      temp_dir <- tempdir()
+      temp_rmd <- tempfile(tmpdir = temp_dir, fileext = ".Rmd")
+      temp_html <- tempfile(tmpdir = temp_dir, fileext = ".md")
+      file.copy(rmd_file, temp_rmd) # Use a copy of the Rmd file to avoid modifying the original
```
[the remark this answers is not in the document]
averissimo: Good point, fixed

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,264 @@
...
+            sprintf("## R Markdown contents are generated from file, please download it from the module UI."),
+            sprintf("# rmarkdown::render(%s, params = params)", shQuote(basename(rmd_file), type = "cmd"))
+          )
+        )
+        out_data@verified <- FALSE # manual change verified status as code is being injected
+      }
+
+      teal.reporter::teal_card(out_data) <- c(
+        teal.reporter::teal_card(out_data),
+        rendered_html_r()
```
gogonzo: markdown fits here better. We can also split it into `list()` suitable with `teal_card()`

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,265 @@
...
+#'       rmd_file = system.file(file.path("sample_files", "test.Rmd"), package = "teal.modules.general"),
+#'       allow_download = FALSE,
+#'       extra_transform = list(nrow_transform)
+#'     )
+#'   )
+#' ) |> shiny::runApp()
+#' @export
+#'
+tm_rmarkdown <- function(label = "RMarkdown Module",
+                         rmd_file,
```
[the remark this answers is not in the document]
gogonzo: > ```r
> #' example_rmd <- paste(
> #'   "---",
> #'   "title: 'Example Document'",
> #'   "output: html_document",
> #'   "---",
> #'   "",
> #'   "## Introduction",
> #'   "",
> #'   "Welcome to this example R Markdown document.",
> #'   "",
> #'   "## Analysis",
> #'   "",
> ```

I think you have a point. Let's not mix two syntaxes. `R` is `R`, `Markdown` is `Markdown` ;] Resolving conversation...

R/tm_rmarkdown.R
gogonzo: > We could keep it as a single markdown object

I think its good to have a single markdown string because anyway at the end (before download) we will concatenate everything to the single file. Only editor will show one combined text-field for this element (i think it is not a big issue so far) 

> the current approach was meant to reduce duplicate rendering

This is a fair reason. I feel (not a strong opinion) that if you render markdown using `rmarkdown::render` and `includeMarkdown` the speed should be similar to `rmarkdown::render` to html. `rmarkdown` renders in two stages to html, first by converting `Rmd -> md` and then pandoc renders from `md -> html`.

The biggest issue with including html is that html can't be converted to pdf, otf etc.

R/tm_rmarkdown.R
averissimo: @llrs-roche parsing and adding the section in `.R` or in `.Rmd` would have similar challenges?

From my initial findings `knitr::purl()` is promising, but it does have some minor limitations with chunk options (it doesn't play well when evaluating params: `{r, eval = !is.null(params$CO2)}`)

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,264 @@
...
+          paste(
+            sep = "\n",
+            sprintf("## R Markdown contents are generated from file, please download it from the module UI."),
+            sprintf("# rmarkdown::render(%s, params = params)", shQuote(basename(rmd_file), type = "cmd"))
+          )
+        )
+        out_data@verified <- FALSE # manual change verified status as code is being injected
+      }
+
+      teal.reporter::teal_card(out_data) <- c(
```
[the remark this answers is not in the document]
averissimo: This has been solved in c563c71f5 by creating a custom `reporter` object for this module.

It keeps the rendered HTML and uses it as caching _(only renders once)_

It also keeps the contents of images in memory to be dumped and used later on. 

Why memory?
- In case files on disk change (multiple reports being added)
- Prevent having to track and keep all images being generated

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,333 @@
...
+    reactive({
+      out_data <- q_r()
+
+      if (allow_download) {
+        out_data <- eval_code(
+          q_r(),
+          paste(
+            sep = "\n",
+            sprintf("## R Markdown contents are generated from file, please download it from the module UI."),
+            sprintf("# rmarkdown::render(%s, params = params)", shQuote(basename(rmd_file), type = "cmd"))
```
gogonzo: `params` not available in the `qenv`

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,333 @@
...
+#' ```yaml
+#' ---
+#' title: "R Markdown Report"
+#' output: html_document
+#' params:
+#'   mtcars: NULL
+#'   iris: NULL
+#' ---
+#' ````
+#'
```
gogonzo: From what I see in the code - `rmarkdown::render(envir = <qenv>)` is used.  
```suggestion
```

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,333 @@
...
+      tryCatch(
+        {
+          rmarkdown::render(
+            temp_rmd,
+            output_format = rmarkdown::md_document(
+              variant = "markdown",
+              standalone = TRUE,
+              dev = "png"
+            ),
+            envir = environment(datasets),
```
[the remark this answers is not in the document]
averissimo: Some functions don't like environments that are `qenv`-based objects. Here is fine though.

Done!

R/tm_rmarkdown.R
averissimo: No longer using param

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,368 @@
...
+
+    q_r <- data_with_output_decorated <- teal::srv_transform_teal_data(
+      "extra_transform",
+      data = pre_decorated_q_r,
+      transformators = extra_transform
+    )
+
+    if (allow_download) {
+      output$download_rmd <- downloadHandler(
+        filename = function() "teal_module.Rmd", # TODO: find a better name
```
osenan: should not have a part of the name date and time?

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,368 @@
...
+  }
+  NextMethod(unclass(x), ...)
+}
+
+#' @describeIn dot-markdown_internal Custom [teal.reporter::to_rmd()] method for `markdown_internal`
+#' object will be used to render the report.
+#' @inheritParams teal.reporter::to_rmd
+#' @param figures_dir (`character(1)`) directory where the R markdown auxiliary files will be saved.
+#' @exportS3Method teal.reporter::to_rmd
+to_rmd.markdown_internal <- function(block, figures_dir = "figures", ...) {
```
osenan: if `markdown_internal` is an internal function shuold not this method be as well?

R/tm_rmarkdown.R
```diff
@@ -0,0 +1,368 @@
...
+    cached_html = rendered_html
+  )
+}
+
+#' @describeIn dot-markdown_internal Custom [tools::toHTML()] method for markdown_internal class that
+#' uses a cached rendering of the module.
+#' @inheritParams tools::toHTML
+#' @param ... Arguments that will be passed to the next method.
+#' @exportS3Method tools::toHTML
+toHTML.markdown_internal <- function(x, ...) {
```
osenan: Should not add @keywords internal as the generic function is internal?

osenan (commented):
Is it possible to create unit tests for `srv_markdown`? With `testServer` I think it has complex enough server logic to implement them.
Also check if other exported funcions we can create uni test.
I tested the app, based on examples. It looks ok:
image
But the markdown text is written by he teal_developr right? Are not the users who can add/remove markdown text

R/tm_rmarkdown.R
averissimo: No, this is true for both `to_rmd` and `toHTML`.

They need to be registered so that `teal.report` can find them when looking to dispatch on `markdown_internal` class.

Otherwise, it won't find the methods registered anywhere and it won't dispatch to these.

Keeping them internal only works if the dispatch call `UseMethod()` would be called from `tmg`

How to check this:
- Remove roxygen tag
- re-generated docs
- restart session
- try the example app

image

image

averissimo:
@osenan I've added tests and I think I addressed your comments.

Let's have a new round of review :-)

R/tm_rmarkdown.R
osenan: I do not know the context very well. It was just strange to me to have the generic function internal but not the methods. But if we require to appear on doc it is fine for me, no need to change.

osenan (approved):
Thanks for the tests. I have run them locally and they pass. They do not pass in CI due to e2e tests.
They are also very useful. There is a lintr error, we could ignore the lint for names in that line.

Issue #929, opened 2025-10-10
[Feature Request]: Update unit test coverage to be at min 80% coverage


donyunardi:
### Feature description

Please make sure unit tests coverage is 80% or more.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

llrs-roche:
We reached the 80% coverage with https://code.roche.com/nest/r-packages/teal.modules.mdr/-/merge_requests/35. 
I'll check if every function has some test coverage

osenan:
[Current coverage](https://github.com/insightsengineering/teal.modules.general/actions/runs/21520234301/job/62009099849) is above 80%

<img width="451" height="509" alt="Image" src="https://github.com/user-attachments/assets/3e718139-5906-4b86-accc-92065590536a" />

PR #961, opened 2026-01-14
Increase coverage for `tm_front_page`

averissimo:
# Pull Request

Part of https://github.com/insightsengineering/teal.modules.general/issues/929

Can't run covr::file_coverage as it tests a deprecated function, but overall it covers 100%

image

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -10,7 +10,6 @@ on:
       - ready_for_review
     branches:
       - main
-      - coverage
   push:
     branches:
       - main


tests/testthat/test-tm_front_page.R
```diff
@@ -0,0 +1,145 @@
...
+      tm_front_page(transformators = list("not a teal_transform_module")),
+      "May only contain the following types: \\{teal_transform_module\\}"
+    )
+  })
+
+  it("show_metadata is deprecated", {
+    testthat::expect_error(tm_front_page(show_metadata = TRUE), "was deprecated")
+  })
+
+  it("accepts valid transformators", {
```
osenan: this last test could be within the module creation `describe` block maybe?

osenan (approved):
The test look ok to me. They are passed and we have 100% of coverage.
I am curious why do we use checkmate functions to verify if is character or string instead of `expect_true(is.character(object))` or similar. It definetely seems more readable actually using checkmate functions.

.github/workflows/check.yaml
```diff
@@ -10,6 +10,7 @@ on:
       - ready_for_review
     branches:
       - main
+      - coverage
```
averissimo:
```suggestion
```

PR #968, opened 2026-01-26
tm_file_viewer tests improvements

m7pr:


osenan:
The two warnings are in tests:
- `handles empty input_path by converting to empty list`
- `handles empty list input_path`

Please add `expect_warning` if this is expected

PR #970, opened 2026-01-27
 Increase coverage for `tm_data_table`

averissimo:
# Pull Request

Part of https://github.com/insightsengineering/teal.modules.general/issues/929

image

### DIFF
--- a/tests/testthat/test-tm_data_table.R
+++ b/tests/testthat/test-tm_data_table.R
@@ -106,10 +106,6 @@ testthat::describe("tm_data_tables input validation", {
       "May only contain the following types: \\{teal_transform_module\\}"
     )
   })
-
-  it("detects deprecated 'datasets_selected' argument", {
-    lifecycle::expect_defunct(tm_data_table(datasets_selected = "ADSL"))
-  })
 })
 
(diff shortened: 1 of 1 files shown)


tests/testthat/test-tm_data_table.R
```diff
@@ -0,0 +1,150 @@
...
+
+  it("creates a module that is not bookmarkable", {
+    testthat::expect_true(attr(tm_data_table(), "teal_bookmarkable", exact = TRUE))
+  })
+})
+
+testthat::describe("tm_data_tables input validation", {
+  it("fails when label is not a string", {
+    testthat::expect_error(tm_data_table(label = 123), "label")
+    testthat::expect_error(tm_data_table(label = c("a", "b")), "label")
```
osenan: minor thing, but this error is because the label has length two? Maybe rename the test tile to
"it fails when label is not a string of the expected length"

averissimo: It tests if it's not a string, so there are 2 expectations, 1 for numeric and another for character vector.

I usually prefer having 1 expectation per test, but due the low complexity of these, I think we can place additional expectations.

If this needs changing then the tests below need as well

PR #973, opened 2026-01-30
Adds a module to support gtsummary

llrs-roche:
Fixes https://github.com/insightsengineering/coredev-tasks/issues/714
Related to https://github.com/insightsengineering/coredev-tasks/issues/676

Adds 3 more package dependencies: (methods), crane, gtsummary, gt. I haven't added minimal version for them.
We could remove crane dependency at the cost of having less corporate style tables (instead of using `crane::tbl_roche_summary()` we could have the vanilla `gtsummary::tbl_summary()`.

Documentation is mostly imported from other packages.

It involves just 3 functions: the module (is this a good name for such module?), the ui and the server function.
Support to download the table depends on https://github.com/insightsengineering/teal.widgets/issues/335. 

Test will come once this branch is reviewed, but known corner case from the module: No `include` variables should select all variables except the one used to split the columns `by`.

Further modifications of the table should be done via decorators not included on this PR.
Adding decorators might involve modifying `check_decorators()` to handle more than one decorator for the same object. This function is duplicated on TMG and TMC and I haven't included on the PR but here is the one I was using:

<details>
<summary>Modified check_decorators to work with multiple decorators for the same object</summary>

```r
check_decorators <- function(x, names = NULL) { # nolint: object_name.

  check_message <- checkmate::check_list(x, names = "named")

  if (!is.null(names) && isTRUE(check_message)) {
    if (length(names(x)) != length(unique(names(x)))) {
      check_message <- sprintf(
        "The `decorators` must contain unique names from these names: %s",
        paste(names, collapse = ", ")
      )
    }
  } else {
    check_message <- sprintf(
      "The `decorators` must be a named list from these names: %s",
      paste(names, collapse = ", ")
    )
  }

  if (!isTRUE(check_message)) {
    return(check_message)
  }

  valid_elements <- vapply(
    x,
    checkmate::test_class,
    classes = "teal_transform_module",
    FUN.VALUE = logical(1L)
  )

  # Nested list
  if (any(!valid_elements)) {
    valid_nested <- vapply(
      x[!valid_elements], function(subdecorators) {
        checks <- vapply(subdecorators,
               checkmate::test_class,
               classes = "teal_transform_module",
               logical(1L))
        all(checks)
      },
      FUN.VALUE = logical(1L)
    )
    if (all(valid_nested)) {
      return(TRUE)
    }
  }

  if (all(valid_elements)) {
    return(TRUE)
  }

[...]

PR #975, opened 2026-02-03
Adds test for gtsummary module

llrs-roche:
# Pull Request

Adds tests for #973

TODO:
- [x] Test input
- [x] Test UI
- [x] Test output

On a different PR:
- [ ] Test decorators:
  - [ ] One
  - [ ] Multiple

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Type: Package
 Package: teal.modules.general
 Title: General Modules for 'teal' Applications
-Version: 0.6.0.9015
-Date: 2026-01-30
+Version: 0.6.0.9016
+Date: 2026-02-05
 Authors@R: c(
     person("Dawid", "Kaledkowski", , "dawid.kaledkowski@roche.com", role = c("aut", "cre")),
     person("Pawel", "Rucki", , "pawel.rucki@roche.com", role = "aut"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# teal.modules.general 0.6.0.9015
+# teal.modules.general 0.6.0.9016
 
 ### Enhancements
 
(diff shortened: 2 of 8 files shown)


R/tm_gtsummary.R
```diff
@@ -96,23 +96,21 @@ tm_gtsummary <- function(
...
-    )
-  }
+  checkmate::assert_string(col_label, null.ok = TRUE)
+  checkmate::assert_list(by, types = "data_extract_spec")
+  assert_single_selection(by)
+  checkmate::assert_list(include, types = "data_extract_spec")
   checkmate::assert_multi_class(pre_output, c("shiny.tag", "shiny.tag.list", "html"), null.ok = TRUE)
   checkmate::assert_multi_class(post_output, c("shiny.tag", "shiny.tag.list", "html"), null.ok = TRUE)
   assert_decorators(decorators, "table")
+  datanames <- teal.transform::get_extract_datanames(list(by = by, include = include))
```
m7pr: Just a note that `get_extract_noun` sounds weird : P
I would either go with `get_noun` or `extract_noun` instead of using 2 verbs with the same name : P

R/tm_gtsummary.R
m7pr: What do you think about that?
In case decorator converts to `gt_tbl` 
```suggestion
    table_r <- reactive({
      out <- req(output_data_decorated())
      # Handle both gtsummary and gt objects
      if (inherits(out[["table"]], "gt_tbl")) {
        out[["table"]]
      } else {
        gtsummary::as_gt(out[["table"]])
      }
    })
```

m7pr: I wonder if `ui_args` grasps `...`?
```suggestion
ui_args <- as.list(environment())
ui_args <- c(ui_args, list(...))
srv_args <- list(...)
```

m7pr (approved):
@llrs-roche if you can resolve that, then I think we can merge

https://github.com/insightsengineering/teal.modules.general/pull/975#discussion_r2763713888

PR #976, opened 2026-02-03
Fix decorators checks

llrs-roche:
# Pull Request

check_decorators didn't raise the error that the decorators provided included something for an object that shouldn't be decorated!

Before:

```r
check_decorators(list(plot = teal::teal_transform_module()), "plot")
## TRUE
check_decorators(list(plot = teal::teal_transform_module()), "table")
## TRUE
```

After:

```r
check_decorators(list(plot = teal::teal_transform_module()), "plot")
## TRUE
check_decorators(list(plot = teal::teal_transform_module()), "table")
## [1] "The `decorators` must be a named list from these names: table."
```

I haven't added tests as the functions are not exported but was found while working on #975 

Once accepted a PR on teal.modules.clinical should be opened with the same changes

### DIFF
--- a/man/tm_g_distribution.Rd
+++ b/man/tm_g_distribution.Rd
@@ -90,8 +90,8 @@ This module generates the following objects, which can be modified in place usin
 \itemize{
 \item \code{histogram_plot} (\code{ggplot})
 \item \code{qq_plot} (\code{ggplot})
-\item \code{summary_table} (\code{datatables})
-\item \code{test_table} (\code{datatables})
+\item \code{summary_table} (\code{DT::datatable})
+\item \code{test_table} (\code{DT::datatable})
 }
 
 A Decorator is applied to the specific output using a named list of \code{teal_transform_module} objects.


llrs-roche:
While checking how this could be introduced, I realized that on the vignette we document tm as:

| Module                 | Output (Class)                                      |
|------------------------|----------------------------------------------------|
| `tm_g_distribution`   | histogram_plot (ggplot), qq_plot (ggplot), summary_table (datatables), test_table (datatables) |


And the module itself accepts these two.

https://github.com/insightsengineering/teal.modules.general/blob/f8d320471cef191076e0b3156b938af25a59cc93/R/tm_g_distribution.R#L1328
https://github.com/insightsengineering/teal.modules.general/blob/f8d320471cef191076e0b3156b938af25a59cc93/R/tm_g_distribution.R#L1335

So I'll fix the assertion on the module itself:

https://github.com/insightsengineering/teal.modules.general/blob/f8d320471cef191076e0b3156b938af25a59cc93/R/tm_g_distribution.R#L197

R/tm_g_distribution.R
```diff
@@ -42,6 +44,9 @@
 #'    decorators = list(
 #'      histogram_plot = teal_transform_module(...), # applied only to `histogram_plot` output
 #'      qq_plot = teal_transform_module(...) # applied only to `qq_plot` output
+#'      summary_table = teal_transform_module(...) # applied only to `summary_table` output
+#'      test_table = teal_transform_module(...) # applied only to `test_table` output
```
m7pr: I think module does not yet support those 2 objects. Check out at those lines

I think in this line

https://github.com/insightsengineering/teal.modules.general/blob/6a0e2fb4a2927959b603258ab9cbeb4fd4d02163/R/tm_g_distribution.R#L1396

we should use `decorated_output_summary_q()` instead of `output_summary_q()`

llrs-roche: Good catch, the `decorated_output_summary_q` was created but not used, `test_table`  is decorated and later used on this line as a few lines above the modules does: `test_q_out <- decorated_output_test_q()`.

Having said, that I tried running the app on the vignette and I couldn't see the UI of the decorator or add the output to the reporter...

m7pr: Yeah, please update then and lets merge

PR #1004, opened 2026-07-07
Restore `gt_summary` and allow for general usage

### LINKED ISSUE #999: Rename `tm_gtsummary` to `tm_tbl_roche_summary`
I learned that there are two main functions in Crane that we use: `tbl_roche_summary` for tables and `tbl_listing` for listings.

I think we will eventually need to support both of these functions.

If `tm_gtsummary` currently runs `tbl_roche_summary` under the hood, what should we call the module that supports `tbl_listing`?

It seems more straightforward to rename `tm_gtsummary` to `tm_tbl_roche_summary`, and then later introduce `tm_tbl_listing`.

averissimo:
# Pull Request

- Fixes #999

### Changes description

- [x] Creates framework to use for `gt`-type data calls
- Implements modules 
    - [x] `tm_gtsummary` with `gtsummary` backend
    - [x] `tm_tbl_roche_summary`
    - [x] `tm_tbl_listing`
- [ ] Discuss solution
- [ ] Discuss removal of `.fun` argument in favor of static implementation only
- [ ] Documentation _(pending technical discussion)_

### Technical details

Creates internal functions called `{tm/srv/ui}_gt_template` which implement a very flexible approach to arbitrary named arguments (also known as dots `...`).

Name dots are dynamically generated because there are no common parameters between target functions.

- **`data_extract_spec`** (future `teal.picks::picks()`): Print them dynamically on encoding panel. It will use data from `teal.data` object
- **`choices_selected`** (future `teal.picsk::variables()`: Print an optional select input on ecoding panel
- Other named dots: Considered static arguments for target function and not rendered in UI

This assumes that any module needs at least 2 implemented functions `tm_<module_name>` `srv_<module_name>_partial` with an optional `ui_<module_name>_partial`

The `partial` allows for custom logic and setting custom and optional UI. 

Note: no need of partial for `crane::tbl_listing` as it has no custom logic. `tbl_summary` and `tbl_roche_summary` need as we're using radio buttons and custom validation.

Below is an example of a partial for `gt_summary::tbl_summary` (must return a qenv-like object to keep with `teal` practices)

```r
srv_gtsummary_partial <- function(id,
                                  data,
                                  .fun_quo,
                                  ...,
                                  summary_args_r) {
    summary_args_processed <- reactive({
      tbl_summary_args <- req(summary_args_r()) # Arguments forwarded from the main server function (template)
      tbl_summary_args$missing <- input$missing # Additional argument from custom UI
      tbl_summary_args$percent <- input$percent # Additional argument from custom UI
      tbl_summary_args
    })

    validated_q <- reactive({ # Custom validation for gtsummary
      q <- req(data())
      summary_args <- req(summary_args_processed())
      validate(
        need(
          length(summary_args$include) != 0L && all(!summary_args$include %in% summary_args$by),
          "Variables to stratify with and variables to include should be different"
        ),
      )
      q
    })

    srv_gt_template_partial(
      id = id,
      data = validated_q,
      .fun_quo = .fun_quo,
      ...,
      summary_args_r = summary_args_processed
    )
  })
}
```

### Example app

Note:

1. Last module uses internal function that works just the same as `tm_gtsummary` (without validation)
    - Shows how flexible this implementation is
1. requires `teal.widgets@999-tm_gtsummary` branch for `tbl_listing` module to work _(it splits the table in 2)_

```r
pkgload::load_all("../teal.widgets")
pkgload::load_all("../teal.modules.general")

add_overall_decorator <- function(col_label = "**Overall**  \nN = {style_number(N)}",
                                  statistic = NULL,
                                  digits = NULL,
                                  last = TRUE,
                                  enable = TRUE) {
  teal_transform_module(
    label = "Overall",
    ui = function(id) {
      ns <- NS(id)
      tags$div(
        bslib::input_switch(
          ns("enable"),
          label = "Enable it?",
          value = enable
        ),
        bslib::input_switch(
          ns("last"),
          label = "Show as last column",
          value = last
        )
      )
    },
    server = function(id, data) {
        teal.logger::log_shiny_input_changes(input, namespace = "teal.modules.gtsummary")
        observeEvent(input$enable, {
          shinyjs::toggleState("last", input$enable)
        })
        reactive({
          if (!input$enable) {
            data()
          } else {
            q <- within(data(),
              {
                table <- gtsummary::add_overall(table,
                  col_label = col_label,
                  statistic = statistic, last = last,
                  digits = digits
                )
              },
              last = input$last,
              statistic = rlang::enexpr(statistic),
              digits = rlang::enexpr(digits),
              col_label = col_label
            )
            validate_qenv(q)
            q
          }
        })
      })
    }
  )
}

data <- within(teal.data::teal_data(), {
  ADSL <- teal.data::rADSL
  ADTTE <- teal.data::rADTTE
})
join_keys(data) <- default_cdisc_join_keys[names(data)]
app <- init(
  data = data,
  modules = modules(
    tm_tbl_roche_summary(label = "Roche Summary (no arguments)"),
    tm_tbl_summary(label = "GT Summary (no arguments)"),
    tm_tbl_roche_summary(
      label = "Roche Summary",
      by = teal.picks::picks(
        datasets(c("ADSL", "ADTTE"), "ADTTE"),
        variables(c("SEX", "COUNTRY", "SITEID", "ACTARM", "CNSR", "PARAMCD"), "SEX")
      ),
      include = teal.picks::picks(
        datasets(c("ADSL", "ADTTE"), "ADSL"),
        variables(c("SITEID", "COUNTRY", "ACTARM", "SEX"), "SITEID", multiple = TRUE)
      ),
      decorators = list(table = add_overall_decorator())
    ),
    tm_tbl_summary(
      label = "GT Summary",
      by = teal.picks::picks(
        datasets(c("ADSL", "ADTTE"), "ADSL"),
        variables(c("SEX", "COUNTRY", "SITEID", "ACTARM"), "SEX")
      ),
      include = teal.picks::picks(
        datasets(c("ADSL", "ADTTE"), "ADSL"),
        variables(c("SITEID", "COUNTRY", "ACTARM"), "SITEID", multiple = TRUE)
      )
    ),
    tm_tbl_listing(
      label = "Table Listing", 
      dataname = "ADSL",
      decorators = list(listing = add_overall_decorator())
    ),
    tm_gt_template(
      label = "GT Template (with gtsumary::tbl_summary)",
      .fun = gtsummary::tbl_summary,
      by = teal.picks::picks(
        datasets("ADSL", "ADSL"),
        variables(c("SEX", "COUNTRY", "SITEID", "ACTARM"))
      ),
      include = teal.picks::picks(
        datasets("ADSL", "ADSL"),
        variables(c("SITEID", "COUNTRY", "ACTARM"), "SITEID", multiple = TRUE)
      ),
      missing = teal.picks::values(c("no", "ifany", "always"), "ifany", multiple = FALSE),
      percent = teal.picks::values(c("column", "row", "cell"), "column", multiple = FALSE)
    )
  )
) |> shiny::runApp()
```


R/tm_gtsummary.R
```diff
@@ -0,0 +1,185 @@
...
+#'   )
+#' )
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+tm_gtsummary <- function(
+  label = "Summary table",
+  by,
+  include,
+  .fun = gtsummary::tbl_summary,
```
llrs-roche:
```suggestion
  .fun = gtsummary::tbl_summary,
```

Shouldn't that be like this? Or we include it because otherwise there is no way to know which UI to use?
With `match.fun` we could check which function is expected and work from there...

R/tm_gt_template.R
```diff
@@ -0,0 +1,246 @@
...
+
+  encodings <- if (length(des_ui) > 0L || length(cs_ui) > 0L || !is.null(partial_ui) || length(decorators > 0L)) {
+    tags$div(
+      tags$label("Encodings", class = "text-primary"),
+      if (length(opts_des) > 0L) teal.transform::datanames_input(opts_des),
+      tagList(!!!des_ui),
+      tagList(!!!cs_ui),
+      partial_ui,
+      # Allow multiple decorators for a single object (table)
+      teal::ui_transform_teal_data(ns("decorator"), transformators = select_decorators(decorators, "table")),
```
llrs-roche: This might be a "name conflict" but for SME team and other internal users a table is different from a listing.
I am not sure if there are some interest from app-developers on having all transformers as argument so that they only have to copy and paste the module and change the specific part of the gt-related function they want.

llrs-roche (commented):
Just two high levels comments

donyunardi:
Hi @averissimo, thanks for your work on this!

I really like where this is going!

First, we have `tbl_gt_template` is the generic engine that can process gt/gtsummary object and can pass any function as an object. 

Then, we have three functions that covers the most common functions using `tbl_gt_template` as the engine.
1. `tm_gtsummary` -> for `gtsummary::tbl_summary`
2. `tm_roche_summary` -> for `crane::tbl_roche_summary`
3. `tm_tbl_listing` -> for `crane::tbl_listing`

For these three functions, since they're dedicated to a specific function, I think the `.fun` argument should be removed and not exposed. I think you eluded to this in your original post as well. 

Also, is it possible to use teal.picks for this? Now that we introduce teal.picks, it make sense to make any new module only teal.picks compatible.

R/tm_tbl_listing.R
```diff
@@ -0,0 +1,85 @@
...
+  dataname = NULL,
+  .fun = crane::tbl_listing,
+  ...,
+  col_label = NULL,
+  pre_output = NULL,
+  post_output = NULL,
+  transformators = list(),
+  decorators = list()
+) {
+  message("Initializing tm_gtsummary")
```
donyunardi:
```suggestion
  message("Initializing tm_tbl_listing")
```

R/tm_gtsummary.R
```diff
@@ -0,0 +1,185 @@
...
+#'           fixed = FALSE
+#'         )
+#'       )
+#'     )
+#'   )
+#' )
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+tm_gtsummary <- function(
```
donyunardi: If this function dedicated only for `gtsummary::tbl_summary`, should we rename it to `tm_tbl_summary` to match the other wrapper functions (i.e. `tm_tbl_listing`)?

R/tm_roche_summary.R
```diff
@@ -0,0 +1,178 @@
...
+#'           fixed = FALSE
+#'         )
+#'       )
+#'     )
+#'   )
+#' )
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+tm_roche_summary <- function(
```
donyunardi:
```suggestion
tm_tbl_roche_summary <- function(
```

Should we rename is to `tm_tbl_roche_summary` to match `tm_tbl_listing`? i.e `tm_<the name of the function>`

R/tm_roche_summary.R
```diff
@@ -0,0 +1,178 @@
...
+
+srv_roche_summary_partial <- function(id,
+                                      data,
+                                      .fun_quo,
+                                      ...,
+                                      summary_args_r) {
+  moduleServer(id, function(input, output, session) {
+    summary_args_processed <- reactive({
+      tbl_summary_args <- req(summary_args_r()) # Arguments forwarded from the main server function (template)
+      tbl_summary_args$missing <- input$missing # Additional argument from custom UI
```
donyunardi:
```suggestion
      tbl_summary_args$nonmissing <- input$nonmissing # Additional argument from custom UI
```

The UI was named `ns(NS("custom", "nonmissing")),` so it should be `input$nonmissing`.

R/tm_gt_template.R
averissimo: Ok, I kept the naming from the previous. I'll change this to listing

R/tm_gtsummary.R
averissimo: It is, that sounds good! I wasn't too strict with the naming as we might need to decide on the convention.

So far we have `tm_t_*` tables, `tm_g_` plots,`tm_a_` and some outliers.

So we should have `tm_tbl_*` to match the calls or use `tm_gt_summary` / `tm_gt_tbl_summary`?

R/tm_gtsummary.R
averissimo: Don't we want to keep the `include` and `by` parameters mandatory?

R/tm_roche_summary.R
averissimo: Sounds good. I wasn't too strict with naming as we need to decide the prefix. See https://github.com/insightsengineering/teal.modules.general/pull/1004/changes#r3551952980

R/tm_roche_summary.R
averissimo: Oh dang, I copy and pasted the whole line and didn't notice it. Thanks

averissimo:
@donyunardi 

> For these three functions, since they're dedicated to a specific function, I think the .fun argument should be removed and not exposed. I think you eluded to this in your original post as well.

I think it's best, the only upside is to allow reuse of same modules if parameters are similar.

> Also, is it possible to use teal.picks for this? Now that we introduce teal.picks, it make sense to make any new module only teal.picks compatible.

I was reworking this today and will push it after this round of reviews :-)

R/tm_gt_template.R
averissimo: Should `tmc` modules have been using this in the `tm_l_**` convention for listings? too late now though.

R/tm_gtsummary.R
llrs-roche: Note that they are not mandatory on the function definitions (or they have defaults). Some people might want to use them or just use the default `= NULL`, for example on `tbl_summary()`, and they are not present on `tbl_listing()`.

R/tm_gt_template.R
llrs-roche: Yes, the templates id use L instead of T:
- Listing: [AEL01](https://insightsengineering.github.io/tlg-catalog/stable/listings/adverse-events/ael01.html)
- Table: [AET01](https://insightsengineering.github.io/tlg-catalog/stable/tables/adverse-events/aet01.html)

But if this was done differently I'm sure there were good reasons or maybe this pattern was not established yet.

Note that I'm saying that depending on what the function used the decorator might be for a table or for a listing. I'm not sure what I used previously or what can be more helpful now.

llrs-roche:
> Then, we have three functions that covers the most common functions using `tbl_gt_template` as the engine.
> 
> 1. `tm_gtsummary` -> for `gtsummary::tbl_summary`
> 2. `tm_roche_summary` -> for `crane::tbl_roche_summary`
> 3. `tm_tbl_listing` -> for `crane::tbl_listing`

@donyunardi and @averissimo, note that some might be interested in some other functions for example `crane::tbl_baseline_chg()`.  

I would keep the flexibility to include other functions. Even if basic functions work reproducing the work of some [functions on crane](https://insightsengineering.github.io/crane/latest-tag/reference/index.html) will be hard. I think we should try to be open to support them. Probably shouldn't support all of them from the start, but we should keep it flexible enough for them.

R/tm_gtsummary.R
averissimo: I would still include them as they have specific validation that you added.

But allowing them to be null as well.

llrs-roche:
We might need to use the machinery to transform NAs to whatever value is needed (NE if I remember correctly from `crane`):

image

The listing should accept a way to `split_by_*` imho.

image


I don't think we should be able to select None on these options:
image

R/tm_gt_template.R
averissimo: Changed all to `listing`, please check it out. Should we rename the name of the modules as well?

Or did you just mean the `tbl_listing`? if that's the case then I'd argue against it as it break from standard :-\

R/tm_roche_summary.R
averissimo: Done in latest

averissimo:
@donyunardi @llrs-roche Updated to picks and incorporated all feedback from PR and discussions.

donyunardi:
Thanks for the update, Andre! Couple comments: 

1. I don't see filter panel for `Roche Summary` and `GT Summary` modules. Why is that?

image


2. Is it possible to choose multiple datasets for `by` and `include` argument in `tm_tbl_roche_summary`?
When I tried to do this, it doesn't seem to like it:
```
    tm_tbl_roche_summary(
      label = "Roche Summary",
      by = teal.picks::picks(
        datasets(c("ADSL","ADAE"), "ADSL")
      ),
      include = teal.picks::picks(
        datasets(c("ADSL","ADAE"), "ADSL")
      )
    )
```

The error message that I got is:
```
Error in teal.picks::is_pick_multiple(by$variables) : 
  Assertion on 'x' failed: Must inherit from class 'pick', but has class 'NULL'.
```

3. in tmg and tmc, we talked about how we want to simplify the module design so that user can just call the module function without any argument and just work. Is it possible to apply the same design here?

llrs-roche:
In addition to Dony's comment, when I try the module `tm_tbl_summary()` I don't see any UI to change the input:

Captura de pantalla 2026-07-14 094743

averissimo:
Thanks for all the feedback team, a bunch of changes happened today and the feature is much more robust.

@donyunardi 

> 1. I don't see filter panel for Roche Summary and GT Summary modules. Why is that?

Filter panel now appears on all modules.

> 2. Is it possible to choose multiple datasets for by and include argument in tm_tbl_roche_summary?
When I tried to do this, it doesn't seem to like it:

That picks is badly formed as it doesn't have a `variables()`, but it revealed 2 things:

1. No check for presence of `variables()` inside the picks in arguments
2. Single dataset support was forced

Both were addressed in the commits today, first with some extra checks and latter with datasets being merged using `picks`.

See updated example with `ADSL`  and `ADTTE`

> 3. in tmg and tmc, we talked about how we want to simplify the module design so that user can just call the module function without any argument and just work. Is it possible to apply the same design here?

Yes, with the exception of `tm_tbl_listing` that requires a `dataname` to indicate which dataset to use

@llrs-roche 

image

R/tm_tbl_roche_summary.R
```diff
@@ -0,0 +1,220 @@
...
+#' )
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+tm_tbl_roche_summary <- function(
+  label = "Summary table",
+  by = NULL,
+  include = teal.picks::picks(
+    teal.picks::datasets(),
+    teal.picks::variables(selected = dplyr::everything(), multiple = TRUE)
```
averissimo: `everything()` might be too much, would it make sense to select just a single column as default?

The first run with everything is very slow, as well as the default UI doesn't show "Select all / deselect all".

Latter problem can be resolved, but I'm worried about it taking too long to run

llrs-roche: It is the default for gtsummary, so I think this is what users would expect. Note that on github there have been some improvements to speed up gtsummary so the demo could be run with that branch too 😉

llrs-roche:
I see the improvements on the UI, I think now it looks great.
As a low hanging fruit maybe we could add decorators so that modifications on the output could be done easily in the demo.

llrs-roche (approved):

averissimo:
Decorators are implemented :-) Please feel free to test it out!

Note: It works fine with decorators, I've updated the example to have one of them.

donyunardi:
Thanks for the update! Looks so good!

Couple findings:

1. Do you see the same thing for table listing?
image

2. For the drop-down, do we not have the button to select / deselect all? I forgot if we did this teal.picks or not. If we didn't, maybe we should add it as an enhancement. 
image

PR #1006, opened 2026-07-22
Implements `teal.picks` picks on `tm_outliers`

averissimo:
# Pull Request

- Part of https://github.com/insightsengineering/NEST-roadmap/issues/36


NAMESPACE
```diff
@@ -25,6 +25,9 @@ S3method(tm_g_scatterplot,default)
 S3method(tm_g_scatterplot,picks)
 S3method(tm_g_scatterplotmatrix,default)
 S3method(tm_g_scatterplotmatrix,picks)
+S3method(tm_outliers,data_extract_spec)
+S3method(tm_outliers,list)
```
llrs-roche: Checking with other modules usually we have a  method for default and a picks one. Shouldn't we change `tm_outliers.data_extract_spec` to `tm_outliers.default`? so that it also works with lists?

R/tm_outliers_picks.R
```diff
@@ -0,0 +1,1172 @@
...
+      obj <- req(data_obj())
+      teal.reporter::teal_card(obj) <- c(
+        teal.reporter::teal_card("# Outliers Module"),
+        teal.reporter::teal_card(obj),
+        teal.reporter::teal_card("## Module's code")
+      )
+      teal.code::eval_code(obj, "library(dplyr);library(tidyr);library(tibble);library(ggplot2)")
+    })
+
+    merged <- teal.picks::merge_srv("merge", data = validated_q, selectors = selectors, output_name = "ANL")
```
llrs-roche, line 310: In other modules after merging there are some other validations and as such it is done before `validate_data()`. 
I can guess why it is not used here but it would be nice if the data was merged before the reactive and the merge output was usedd for validation.

R/tm_outliers_picks.R
```diff
@@ -0,0 +1,1172 @@
...
+          length(categorical_var) == 0,
+          ANL,
+          ANL[, names(ANL) != categorical_var, drop = FALSE]
+        ),
+        min_nrow = 10,
+        complete = TRUE,
+        allow_inf = FALSE
+      )
+      validate(need(is.numeric(ANL[[outlier_var]]), "`Variable` is not numeric"))
+      validate(need(length(unique(ANL[[outlier_var]])) > 1, "Variable has no variation, i.e. only one unique value"))
```
llrs-roche: Indeed, here are some validations that might be better on the previous reactive.

R/tm_outliers_picks.R
```diff
@@ -0,0 +1,1172 @@
...
+      outlier_definition_param <- if (method == "IQR") {
+        input$iqr_slider
+      } else if (method == "Z-score") {
+        input$zscore_slider
+      } else if (method == "Percentile") {
+        input$percentile_slider
+      }
+
+      # this is utils function that converts a %>% NULL %>% b into a %>% b
+      remove_pipe_null <- function(x) {
```
llrs-roche: This is silly but if this function is now defined in two different places mabye it should be on utils.R and be reused accross modules.

R/tm_outliers_picks.R
```diff
@@ -0,0 +1,1172 @@
...
+              substitute(dplyr::ungroup())
+            }
+          )
+        ) %>%
+          remove_pipe_null()
+      )
+
+      # ANL_OUTLIER_EXTENDED is the base table
+      dataname_first <- outlier_dataname()
+      join_keys <- as.character(teal.data::join_keys(data_obj())[dataname_first, dataname_first])
```
llrs-roche, line 458: If we use `merge_srv` shouldn't the join keys already be addressed by that function? 

Note that merge_srv can use any data and join `ANL_OUTLIER` with `dataname_first` too. Just commenting to avoid duplicating code (but keep in mind the drawbacks too).

llrs-roche (commented):
I like that now examples are much shorter. Nice work.

Just a few comments

NAMESPACE
averissimo: Good point, I'll convert this to be identical with others

R/tm_outliers_picks.R
averissimo: Good point. it will be removed once the old code is, but that will take a while

R/tm_outliers_picks.R
averissimo: It does have a `validated_q` before the merge, which performs validations that don't require data inspection.

That has been the standard, right? (screenshot from swimlane in osprey)

image

R/tm_outliers_picks.R
averissimo: These use the merged data, so it wasn't validated in `validate_q`, but it would fit there!

I'll move it as it doesn't have any more complex logic or conditions

R/tm_outliers_picks.R
averissimo: I didn't touch the logic in the module, but good eye to catch this

However, this is another instance of the need to keep all the variables when merging.

It is used to add additional columns in the summary table on the bottom:

image

averissimo:
@llrs-roche I think I addressed the comments from the previous review

R/tm_outliers_picks.R
llrs-roche: Mmh, maybe having just input validation for `validated_q` is the standard. I thought it was to split the logic of the module vs the input (even after merging it) checking. 

But I'll mark this  as resolved.

R/tm_outliers_picks.R
```diff
@@ -346,12 +357,13 @@ srv_outliers.picks <- function(id, # nolint: object_name_linter.
           )
         }
       } else {
-        validate(need(
-          is.factor(ANL[[categorical_var]]) ||
+        teal::validate_input(
+          inputId = "categorical_var-variables-selected",
```
llrs-roche, line 361: Sorry to add a new comment, but wouldn't it be better to use: `inputId = "categorical_var"`?

llrs-roche (approved):
Looks great. just one last additional comment

R/tm_outliers_picks.R
averissimo: Unfortunately not, as that inputId doesn't exist nor it is smart enough to try "-variables/-values/-datasets"

The validation is generic by design as it lives on teal without any direct knowledge of picks

donyunardi: Since this is minor, I will resolve and merge for now so we can move forward with the release.

